PR #53, opened 2015-01-19
Add link to burn your select tags

gemmaleigh:
The best way to explain why we’d like people not to use native select
tags is to get them to watch this talk.

Also mention the HTML select element (next to drop-down lists), so it’s clear we’re talking
about the same thing.

### DIFF
--- a/views/index.html
+++ b/views/index.html
@@ -773,7 +773,10 @@
 
     <h3 class="heading-medium" id="form-select-boxes">Drop-down lists</h3>
     <p class="text">
-      Avoid using drop-down lists - use radio buttons or checkboxes instead.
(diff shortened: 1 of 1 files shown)


tombye:
I agree this is a great way to illustrate the evidence behind that point, just to say it's worth checking with the author before publishing this. 

@alicebartlett are you ok with this forming part of elements?

gemmaleigh:
I should have asked @alicebartlett in the first place, sorry Alice and let me know if this is okay with you.

I'm trying to provide as much evidence from user research as I can behind our guidance, hence wanting to add this link - and the other links to relevant hackpads too.

PR #57, opened 2015-01-28
Improve details summary polyfill

gemmaleigh:
The current polyfill wasn't correctly detecting the "open" attribute and was firing twice in Chrome, 
when the summary was clicked.

For now, it has two checks - first for native support, which uses the boolean "open" attribute.  Then for non-native support, which checks for an "open" attribute with an empty value, otherwise "open" is
undefined (not present) and the details content should be closed.

For non-native support, an open or closed arrow (pointing right or down) is added, and styled to look like the native arrow in Chrome - depending on the initial "open" state. This arrow is hidden from assistive technology using `role="presentation"`. [Read the W3C Recommendation here](http://www.w3.org/TR/wai-aria/roles#presentation).

### A detailed explanation from Léonie:

`<details role="group">`
 
The implicit role for `<details>` should be "group". Since most browsers haven't implemented support for the element (let alone accessibility support), we need to apply this role manually – so that assistive technologies know that it's a container for a group of related content.

`<summary role="button" aria-controls="details-content-0" tabindex="0" aria-expanded="false">…</summary>`

#### [1] role="button * 

The implicit role for `<summary>` should be "button", but as with `<details>` we need to provide the accessibility support explicitly using the role attribute.

#### [2] aria-controls="details-content-0" 

* Note: `aria-controls` is not consistently supported by assistive technologies. 
Those that do support it usually provide a keyboard shortcut that lets the user move focus to the start of the element being controlled. It is not expected behaviour for a screen reader to automatically start to speak the content of the disclosure widget when it opens. 
* It's also worth noting that VoiceOver/Safari doesn't support `aria-controls` in any case.

#### [3] tabindex="0" *

Without basic browser support for `<summary>` it can't be focused on with the keyboard. Using tabindex="0" places the element into the natural tab order of the page, so as someone using a keyboard tabs through the page they'll be able to focus on it in order to activate it.

#### [4] aria-expanded="false" *

The aria-expanded attribute informs assistive technologies that the `<summary>` element triggers the expansion/collapse of something. When focus first moves to the `<summary>` element, its state is announced as "collapsed", and once the button is activated it changes to "expanded".

### DIFF
--- a/public/javascripts/vendor/details.polyfill.js
+++ b/public/javascripts/vendor/details.polyfill.js
@@ -103,7 +103,6 @@
 
         details.__summary.setAttribute('aria-expanded', 'true');
         details.__content.setAttribute('aria-hidden', 'false');
-
         details.__content.style.display = 'block';
       }
 
@@ -158,14 +157,13 @@
         if (details.getAttribute('open') === "") {
           twisty.className = 'arrow arrow-open';
           twisty.appendChild(document.createTextNode('\u25bc'));
-          details.__summary.__twisty = details.__summary.insertBefore(twisty, details.__summary.firstChild);
-
         } else {
           twisty.className = 'arrow arrow-closed';
           twisty.appendChild(document.createTextNode('\u25ba'));
-          details.__summary.__twisty = details.__summary.insertBefore(twisty, details.__summary.firstChild);
-
         }
+
+        details.__summary.__twisty = details.__summary.insertBefore(twisty, details.__summary.firstChild);
+        details.__summary.__twisty.setAttribute('aria-hidden', 'true');
 
       }
     }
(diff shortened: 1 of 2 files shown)


edwardhorsford:
Really interesting summary. Thanks.

tombye:
This is a great improvement and the summary of why the changes are needed is brilliant. :+1:

tombye:
I've re-read the spec and my original presumption was incorrect so it's worth explaining why, mainly because it also explains why using `aria-hidden` is actually the advised approach.

In our use-case, the markup we tried for the arrow is:

```
<i class="arrow" role="presentation">▼</i>
```

The ▼ is a text node which is a child of the `<i>` tag. The spec has this to say about how to handle descendants of elements given the role of 'presentation':

"...the user agent MUST expose content and descendant elements that do not have an explicit or inherited role of presentation."

The ▼ text node has a role of "text" so it will be exposed. This explains why the arrow was still being spoken when the `<i>` tag was given the 'presentation' role.

PR #59, opened 2015-02-05
escape square brackets in element name

stephenrichards:
Rails creates form elements with square brackets, for example customer[name].  These need to be escaped in application.js showHideToggledContent()

### DIFF
--- a/public/javascripts/application.js
+++ b/public/javascripts/application.js
@@ -1,5 +1,11 @@
 function ShowHideContent() {
   var self = this;
+
+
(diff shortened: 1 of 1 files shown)


tombye:
Hi Stephen, thanks for the pull request. Quite nice to have it explicit that some extra processing may be required for certain implementations.

@gemmaleigh are you happy for this to go in?

gemmaleigh:
@tombye @stephenrichards LGTM, thanks!

PR #62, opened 2015-03-10
Add validation guidance and examples

gemmaleigh:
Validation styles and guidance, plus examples.

This can be previewed here: http://govuk-elements-test.herokuapp.com/#guide-errors
with static error examples on the index page and three example pages, following Léonie's advice.

[1. A single question per page]
(http://govuk-elements-test.herokuapp.com/examples/form-validation-single-question)
[2. A single question per page - radio buttons]
(http://govuk-elements-test.herokuapp.com/examples/form-validation-single-question-radio)
[3. Multiple questions]
(http://govuk-elements-test.herokuapp.com/examples/form-validation-multiple-questions)

gov uk elements
gov uk elements-2
gov uk elements-3

 
gov uk the best place to find government services and information
gov uk the best place to find government services and information

### DIFF
--- a/app.js
+++ b/app.js
@@ -12,6 +12,8 @@ app.set('views', __dirname + '/views');
 // Middleware to serve static assets
 app.use('/public', express.static(__dirname + '/public'));
 app.use('/public', express.static(__dirname + '/govuk/public'));
+app.use(express.json());       // to support JSON-encoded bodies
+app.use(express.urlencoded()); // to support URL-encoded bodies
 
 // routes (found in routes.js)
 
--- a/govuk/public/javascripts/govuk/analytics/error-tracking.js
+++ b/govuk/public/javascripts/govuk/analytics/error-tracking.js
@@ -1,0 +1,27 @@
+// Extension to track errors using google analytics as a data store.
+(function() {
+
+  "use strict";
+
+  GOVUK.analyticsPlugins = GOVUK.analyticsPlugins || {};
+
+  GOVUK.analyticsPlugins.error = function () {
+    var trackJavaScriptError = function (e) {
+      var errorSource = e.filename + ': ' + e.lineno;
+      GOVUK.analytics.trackEvent('JavaScript Error', e.message, {
+        label: errorSource,
+        value: 1,
+        nonInteraction: true
+      });
+    };
+
+    if (window.addEventListener) {
+      window.addEventListener('error', trackJavaScriptError, false);
+    } else if (window.attachEvent) {
+      window.attachEvent('onerror', trackJavaScriptError);
+    } else {
+      window.onerror = trackJavaScriptError;
+    }
+  }
+
+}());
--- a/govuk/public/javascripts/govuk/analytics/google-analytics-classic-tracker.js
+++ b/govuk/public/javascripts/govuk/analytics/google-analytics-classic-tracker.js
@@ -1,0 +1,111 @@
+(function() {
+  "use strict";
+  window.GOVUK = window.GOVUK || {};
+
+  var GoogleAnalyticsClassicTracker = function(id, cookieDomain) {
+    window._gaq = window._gaq || [];
+    configureProfile(id, cookieDomain);
+    allowCrossDomainTracking();
+    anonymizeIp();
+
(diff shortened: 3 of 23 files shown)


edwardhorsford:
Interesting. Like the extra weight of colouring the question title. Do we really need the duplicate 'please answer x' as well though?

What about nuances like Register to vote's, that put the validation message near where you click (client side validation, so page refresh not required).

gemmaleigh:
When there's only one question per page and we're not showing a summary box with a heading "There was a problem submitting the form", yes we do. I'm keen to have a statement indicating something has gone wrong "Please answer x" or "Enter your x" (in addition to the bold label and border), to ensure this is read out for anyone using Assistive Tech. 

[These examples are based on feedback from Léonie](https://designpatterns.hackpad.com/Error-messaging-and-validation-HiGXPZ7XiG0) on how we should be marking up our forms. I'll ask @joelanman and @tombye for their thoughts on Register to Vote.

joelanman:
Validation messages are traditionally placed high on the page, because validation required a full page refresh. This means a user would be looking at the top of the page, and the validation message explains why they have not continued to the next page.

Register to vote uses JavaScript validation for a faster, more fluid response to user input. The idea of placing the validation message just above the continue button is that in this situation, that is where the user's eye-line is. In general I don't like jumping people around a page as it can be disorientating.

Maybe for now it would be enough to say that the styles/layouts in this PR relate to server-side validation?

edwardhorsford:
Also that there's a need to be pragmatic with the design - the aim of validation is ultimately to let a user know that something didn't work, and steps to fix. Different situations have different needs.

tombye:
I agree with Joe that part of this pattern is designed around a page reload (the summary appearing at the top) should be marked as such. We advise that sites should use progressive enhancement so these patterns will always be implemented at the HTML/CSS level.

We could probably do with more testing around how JavaScript validation layers on top of this (ie. how fall-through is handled and what users read from the difference in messaging speed) before looking at that in this level of detail.

jonnywyatt:
Should required and aria-required be added to compulsory inputs? Can suppress any native validation with "novalidate" attr on the \<form\>

edwardhorsford:
We suppress native validation on passports. Better to have a single consistent way of doing it (in my view).

edwardhorsford:
One case I want to do better on is where there's only a single field on the page, but it's below a lot of content. We have a declaration style page on passports like this, and as the field is 'below the fold', the validation message at the top is out of context for users and confuses them.

tombye:
@jonnywyatt We have a pattern to surface which fields are required through the text in their labels so that information is available to users at the point they decide which fields to enter something into:

https://designpatterns.hackpad.com/Required-and-optional-fields-Vl1ydPMta7T

I'd argue that use of native validation (through `aria-required` & `novalidate`) is to do with the JavaScript layer which is out of the scope of this particular pull request. No reason why it can't form part of further advice in future where it can be covered properly.

tombye:
@edwardhorsford I don't think your example runs contrary to this pattern as long as the reasoning for having a summary or not is made clear. 

In the majority of cases having no summary for pages with a single field means less content to navigate so is the right thing to do but that reasoning doesn't discount the need you describe. It would be useful if you could add that example to the hackpad so everyone is aware of the possible variations when checking the evidence for that decision.

jonnywyatt:
@tombye - aria-required would be for screenreaders rather than validation, but see what you're saying, if the labels indicate that then job done.

tombye:
@jonnywyatt yeah, that's a better explanation of what I meant: that information should sit at the level most users can access which is always text above semantics.

edwardhorsford:
@tombye in that case we have a summary, but it still isn't sufficient for all users. I'm thinking of ways to improve this. Unrelated to this PR though.

kelliematheson:
@gemmaleigh Will these new screenshot examples be added to the elements page?

gemmaleigh:
Hi @kelliematheson,

The latest work on errors and validation is here: 
https://github.com/alphagov/govuk_elements/pull/64

There will be working examples, rather than screenshots 
- the screenshots are to preview the changes to be made.

PR #63, opened 2015-04-09
Full width buttons for mobile

gemmaleigh:
- Remove button styles from _forms.scss
- Set all buttons to be 100% for mobile (640px and below)


gemmaleigh:
Also a PR for the govuk prototype kit: https://github.com/alphagov/govuk_prototype_kit/pull/6

PR #64, opened 2015-04-10
New errors and validation guidance

gemmaleigh:
Validation styles and guidance, plus examples.

This can be previewed here: http://govuk-elements-test.herokuapp.com/#guide-errors
with static error examples on the index page and three example pages, following Léonie's advice.


cjforms:
Looks good to me



Caroline Jarrett
User researcher / forms specialist
Government Digital Service

Twitter @cjforms
07990 570647

My next GDS days
Mon 30 Mar: Aviation House
Tue 31 Mar: Aviation House
Wed 1 Apr: Aviation House

Thu 8 Apr: Working at home
Fri 9 Apr: Working at home

matthewford:
Should the error message not go between the title  and field?

cjforms:
It doesn't matter that much on an extremely easy question, but those
questions are far less likely to provoke errors in the first place.

If the question is complex, then the relationship between the various parts
of the question (heading, label, examples, links to further help)  is also
likely to be more complex. It becomes harder to say exactly where to insert
the error message, and putting the error message into the middle of all of
that may separate them from each other - visually or conceptually - and
make it harder for users to understand exactly what the error message means
and is relating to.

If we say 'put the error message ahead of the title', it works both for the
questions where it's much better and for the questions where it doesn't
matter all that much. This means we take out an 'it depends' and we don't
have designers worrying about whether or not a particular question "counts
as" complex or not.

Gemma chose simple examples to illustrate placement, and that works for me
because we want to keep this advice simple and avoid complicated questions
that might need a backstory to explain them.

Does that all make sense?

Caroline

matthewford:
Sure. However to counter that I would say that complex examples is where we *do* need more examples/guidance. 

Even in this simpler example, if the error message isn't near the field in question, users could confuse which field the error is related to as it's close to the bottom of the preceding field. I know the border should help with this, but say you have more than one line of help text I could see it cause some issues.

cjforms:
Sure. We're discussing the problem of how to write the error messages on
the hackpad. Please join in on that discussion there.

To give some context about this pull request: its primary aims are
1. to ensure that the error messages are accessible for people who use
screen readers, and also work for everyone else too
2. to be an improvement to the existing rather minimal advice.

No doubt we'll be thinking some more about this advice once we've got our
ideas sorted out on writing the error messages.
Meanwhile, this pull request moves us forward in helpful ways.

Best
Caroline



Caroline Jarrett
User researcher / forms specialist
Government Digital Service

Twitter @cjforms
07990 570647

My next GDS days
Mon 13 Apr: Aviation House
Tue 14 Apr: Aviation House
Wed 15 Apr: Aviation House

Mon 21 Apr: IOE workshop (am) Aviation House (pm)
Tue 22 Apr: Aviation House
Wed 23 Apr: Aviation House

matthewford:
@gemmaleigh this pr cannot be merged via the web, it needs to be rebased against the current master branch first

gemmaleigh:
Hi @matthewford, it's not ready to be published yet, but thanks for letting me know.

saproctor:
A couple of points about the summary box.

The title is coded as a H3 that is placed before the H1 on the screen. Does this not infringe the WCAG 2.4.10 principle about using headings to convey a logical hierachy?

The line in the summary box "Description of the errors and how to correct them". The errors should do both of these things so this line is unnecessary. I think there should be a line of introduction to set context.

The errors in the summary box are coded like a bullet pointed list (<ul>) but don't have visible bullet points. The GDS style guide says bullets should have a lead-in sentence and a lower case letter at the start. I think a lead-in sentence would help set context for the errors.

cjforms:
On the WCAG 2.4.10 principle: I'll leave it to Gemma to answer on the coding points. But even as coded, the method doens't contravene the principle because the user can only get to this amended page after encountering the previous page where the H1 is the most important level. So what we're doing with the error message is adding a commentary to the page - does that help?

On your point about 'description of the errors and how to correct them': as I mentioned earlier, these very simple examples are designed to show placement without confusing the issue with complexity in the examples, In practice, the questions can be highly complex and users may need additional information on how to correct the error. If they don't need the additional information then it's fine to leave it out.

saproctor:
I agree that wee're adding a commentary to the page, but by making the H3 the most superior heading on the screen, a screen reader could still read the entire page with the commentary so the H3 would supercede the H1. We could place it underneath the H1, code it as a H2 and that would satisfy the WCAG.

What do you think about displaying the bullet points?

cjforms:
OK, fair point. I'll leave it to Gemma to take a view on it. 

Also on the bullet points.

gemmaleigh:
Hi @saproctor,

The most important thing in these examples is that the summary box sits at the top of the main content area. This ensures it will be read out by Assistive Technology before the rest of the page content.

The argument over heading levels is really missing the point here. The summary box needs to be associated with a heading, to ensure the heading is read out immediately when focus is set to the summary box. 

The heading could be any level, here it is at the top of the page so it may well be more appropriate to use a H1. I'll test this using a screenreader, to find out how this heading level affects the heading hierarchy before making a recommendation.

As for the bullet points, this is ultimately @markhurrell's decision - he's Design Lead for GOV.UK.
There is no requirement for this list of links to have a lead-in sentence or to use bullet points. 

It should be as fast as possible to find the errors that have occurred and correct them. If this can be achieved by having well-written links in the summary box, without a decorative bullet point - that is ideal.

saproctor:
Morning @gemmaleigh,

The reason why I mention the headings is because I know service that have failed accessibility reviews for breaking heading logic, including there being more than one H1, and H2s and H3s appearing before the H1.

I mentioned the bullets being visible because (1) they are coded as an unordered list, and (2) the style guide actually says "Lists should be bulleted to make them easier to read."

gemmaleigh:
Closing this PR.
There is a new branch with a clearer version of this work here:
https://github.com/alphagov/govuk_elements/tree/new-errors-and-validation-guidance-for-review
This will shortly become a new pull request to be reviewed.

PR #71, opened 2015-05-14
New errors and validation guidance for review

gemmaleigh:
There has been rather a lot of back-and-forth in the previous PR while this was tested further.
This is a new clean branch with the proposed changes to the errors and validation section.

It can be previewed here: https://govuk-elements-test.herokuapp.com/

Our current thinking about validation messages is:
- encourage the use of error summary boxes, as close to the top of the main content as possible
- provide a link to correct each error within the summary box
- error text must sit within either the label or legend for the question it relates to, to ensure it is read out by Assistive Technology

There are two example pages, one which shows how to display errors for a single question - a group of radio buttons and another which shows how to display errors for multiple questions. They have additional implementation advice underneath each example.

In both cases a summary box has been used and the wording is generic, to discourage these examples being copied verbatim. The words used as error messages in the examples show how the messages should differ, however, the exact wording used for error messages is outside the scope of this pull request.

gov uk elements - errors

gov uk the best place to find government services and information - example 1
gov uk the best place to find government services and information - example 3

### DIFF
--- a/views/snippets/form_error_multiple_show_errors.html
+++ b/views/snippets/form_error_multiple_show_errors.html
@@ -1,0 +1,59 @@
+<div class="error-summary" role="group" aria-labelledby="error-summary-heading-example-2" tabindex="-1">
+
+  <h1 class="heading-medium error-summary-heading" id="error-summary-heading-example-2">
+    Message to alert the user to a problem goes here
+  </h1>
+
+  <p>
+    Optional description of the error(s) and how to correct them
+  </p>
+
+  <ul class="error-summary-list">
+    {{^fullName}}
+    <li><a href="#error-full-name">Descriptive link to the question with an error</a></li>
+    {{/fullName}}
+    {{^niNo}}
+    <li><a href="#error-ni-number">Descriptive link to the question with an error</a></li>
+    {{/niNo}}
+  </ul>
+
+</div>
+
+<h1 class="heading-large form-title">
+  Your details
+</h1>
+
+<div class="form-group error">
+
+  <label class="form-label-bold" for="error-example-full-name">
+
+    Full name
+
+    <span class="form-hint">As shown on your birth certificate or passport</span>
+    <span class="error-message" id="error-full-name">Error message about full name goes here</span>
+
+  </label>
+
+  <input class="form-control" id="error-example-full-name" type="text" name="fullName" value="" aria-describedby="error-full-name">
+</div>
+
+<div class="form-group error">
+
+  <label class="form-label-bold" for="error-example-ni-number">
+
+    National Insurance number
+
+    <span class="form-hint">
+      It's on your National Insurance card, benefit letter, payslip or P60.
+      <br>
+      For example, 'VO 12 34 56 D'.
+    </span>
+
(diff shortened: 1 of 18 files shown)


edwardhorsford:
Gemma - these are excellent. After using these on passports for our last user research session, I'm pleased. The warnings are very clear.

Would be good to follow up at some point with validation for inset / conditional questions.

saproctor:
A few questions...

1. Are we saying there should be a summary box on all screens - even those with a single question? What would this mean when the anchor is already visible on the screen?

2. Should the summary box outline be the full measure of the wrapper? None of its content will be.

3. Can we have two H1s? Whether it is against WCAG or not, how does it affect users that use the headings to navigate the page?

4. Would having a <ul> in the summary box that has hidden bullet points and a capital letter at the beginning infringe GOV.UK list style? Or would this be an accepted style for summary boxes?

views/snippets/form_error_multiple_show_errors.html
```diff
@@ -0,0 +1,59 @@
...
+</h1>
+
+<div class="form-group error">
+
+  <label class="form-label-bold" for="error-example-full-name">
+
+    Full name
+
+    <span class="form-hint">As shown on your birth certificate or passport</span>
+    <span class="error-message" id="error-full-name">Error message about full name goes here</span>
```
gavboulton: You're linking to the input text error message here, but for the radio button example you've linked to the legend, https://github.com/alphagov/govuk_elements/blob/new-errors-and-validation-guidance-for-review/views/snippets/form_error_radio.html#L31. Is this intentional?

gemmaleigh: Hi @gavboulton,

No, that was a mistake.

I've added the missing aria-describedby attributes and the link is now (and should be) to the label, in the same way as the radio button example links to the legend.

gavboulton: Hi @gemmaleigh,
This is looking really good.

robmckinnon:
@gemmaleigh When are you expecting this PR to be merged?

We are keen to re-use your new errors css in [ministryofjustice/defence-request-service](https://github.com/ministryofjustice/defence-request-service).

gemmaleigh:
Hi @robmckinnon, apologies - I've been away. It'll be merged later today.

saproctor:
@gemmaleigh 

The descriptive link for an incorrect National Insurance number in the example with multiple error messages doesn't work.

The link points at an id (error-ni-number) that either isn't there (only one error is made) or incorrect (error-full-name when there are two errors).

gemmaleigh:
@saproctor thanks for pointing that out, those examples have now been updated.

gemmaleigh:
@saproctor

1. Yes, this is the best method we've found for users of Assistive Technology.

2. The summary box should be the full width of its container.

3. If it is found that using headings in this way causes a problem, the examples will be revised. For now, they reflect that the most important heading on the page - once an error has been made - is the heading in the summary box.

4. Yes, this is the agreed style for summary boxes. I've spoken to Beck Thompson and the Content Style Guide will be updated to note the differences for lists within summary boxes.

PR #74, opened 2015-05-29
Only set $path if not already set

ztolley:
Some users of the GOV UK Elements sass have their own existing file structure for image assets. When the $path was added it created urls starting with /public/images but it would be more flexible to allow other users to set their own value. By making the assignment optional, if a developer has defined $path in their top level file it will use that instead. Alternatively move the $path assignment out of the helper and up to the main.scss file

### DIFF
--- a/public/sass/elements/_helpers.scss
+++ b/public/sass/elements/_helpers.scss
@@ -7,8 +7,10 @@
 @import "typography";
 @import "css3";
 
-// Path to assets for use with file-url()
-$path: "/public/images/";
+// Path to assets for use with file-url() is not already defined
+@if ($path == false) {
+    $path: "/public/images/";
+}
 
 // Return ems from a pixel value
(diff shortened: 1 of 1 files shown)


gemmaleigh:
As `_helpers.scss` shows which partials are imported from the govuk front end toolkit, I'd like to keep the $path variable here, rather than moving it up into `main.scss`.

Ideally, I'd set a `_vars.scss` file above this and keep all Sass variables the same place, however, all other variables used are being set by the front end toolkit (for this project).

I've added the missing import to make it clear [_url-helpers.scss](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/_url-helpers.scss) is a dependency. 

I'll merge this so it doesn't interfere if $path is already declared, apologies for the slow response on this one - I was away last week.

PR #76, opened 2015-06-11
Ignore top margin of element following a legend in Webkit

timpaul:
In Webkit, the bottom margin of a legend and the top margin of the next element don't overlap, which they should. This addresses that by ignoring the top margin of the next element. As an approach this feels a bit fragile though...


gemmaleigh:
I'd prefer not to introduce fixes for specific browsers, this issue was closed with Normalize. https://github.com/necolas/normalize.css/issues/168

Closing for now.

PR #77, opened 2015-06-11
Set textarea to block so Chrome and FF don't add a bottom margin

timpaul:


gemmaleigh:
This feels like it should sit in a reset file.
Is there any reason not to set margin:0 for `<textarea>` by default?

Normalize.css just removes the margin here.
https://necolas.github.io/normalize.css/3.0.2/normalize.css

/**
 * 3. Address margins set differently in Firefox 4+, Safari, and Chrome.
 */

button,
input,
optgroup,
select,
textarea {
  margin: 0; /* 3 */
}

PR #78, opened 2015-06-11
Add missing form control widths

timpaul:
Adding these allows us to set form control widths equal to all the main grid widths. I've also set the two smallest widths to work like the others and flip to 100% at mobile. 12.5% width at mobile width may well be too small for the intended input. If there's a known character width people can use those styles instead (see pull request #81).

Not sure if .form-control-1-1 is the right way to specify 100% width. We need it though, for when form controls are used in columns.


gemmaleigh:
Hi @timpaul, is this any better?

By default, `.form-control` is set to 100% width, to fill the full width of the grid column it sits in.

This will remove the need for `.form-control-1-1` and will mean a `.form-width` class only needs to be set for widths smaller than 100%.

timpaul:
Yeah - I prefer that as well. Does this constitute a breaking change? In that it's going to ping a lot of people's fields from 50% to 100%?.

edwardhorsford:
I don't understand the use case for these.

In nearly all cases we know the expected length of the data users will be provided. Our own guidance says the boxes should be sized to suggest the correct length. Thus boxes in general should have a fixed width, not percentage based off of the grid.

My preference would be to multiples of the gutter margin or something. so 1x, 2x, 4x, 8x, 16x (60px, 60px, 120px, 240px, 480px, etc)

gemmaleigh:
Closing this PR in favour of #86.
The missing widths were added in #85 and will be replaced in #86.

Please add comments there.

PR #79, opened 2015-06-11
Remove top-margin from 1st element in indented panel

timpaul:
If your indented panel start with a heading (or another element with a top-margin) you end up with unwanted whitespace. This fixes that.


gemmaleigh:
This has been fixed here:
https://github.com/alphagov/govuk_elements/blob/master/public/sass/elements/_panels.scss#L13

Please can you close this, @timpaul, if it works for you.

PR #80, opened 2015-06-11
Add styles for horizontal rules

timpaul:
If people do use them, we may as well make them look good.

### DIFF
--- a/public/sass/elements/_elements-typography.scss
+++ b/public/sass/elements/_elements-typography.scss
@@ -190,12 +190,15 @@ code {
 }
 
 // Horizontal rule style
-hr{
+hr {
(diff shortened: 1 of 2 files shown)


gemmaleigh:
LGTM - I'd like to add an example. 
Can you show me somewhere they've been used?

gemmaleigh:
Since it should be used for a thematic break between paragraphs, I've added two paragraphs to the typography example page.

For any other dividing lines, a div with border would be a better solution.

---
http://html5doctor.com/small-hr-element/
The hr element represents a paragraph-level thematic break.
[W3C specification](http://w3c.github.io/html-reference/hr.html)

PR #81, opened 2015-06-11
Add fixed character width form controls

timpaul:
For short form fields where the maximum number of characters is known it can be more useful to have fixed widths. These styles allow for that.

Note: the widths may need tweaking and there's probably a neater way of specifying these styles in sass.


gemmaleigh:
This will form part of the fixed-width form widths PR, so closing for now.

PR #82, opened 2015-06-11
Inline form-groups for compound controls

timpaul:
There's a need to be able to align form controls horizontally sometimes. The date-of-birth picker is one example, but there are others (units-per-time-period for example).

Rather than have a date-specific style, why not have a generic 'inline-form-groups' wrapper?

That's what this pull request does. As it stands, I've removed the old date styles, so it's a breaking change for anyone who's used them.

Also, this request relies on #81 (fixed width form controls) - but I added it again as a commit here (sorry).


gemmaleigh:
I've moved the styles for generic inline-form-groups over to:
https://github.com/alphagov/govuk_elements/tree/add-test-form-page

Rather than make this a breaking change, I'll leave the form-date pattern - and mark it as deprecated, 
this will mean it won't break for anyone using it - but we can let them know to update the html to use the new pattern. I'll also update the examples so they show how to use the new inline wrapper.

PR #85, opened 2015-07-07
Make forms sass clearer and propose changes

gemmaleigh:
This pull request:

- Adds a numbered list of contents
- Numbers individual sections
- Marks styles which are likely to change with TODO: and the proposed change.

Most of these proposed changes exist in other pull requests (by @timpaul).

This is a tidy-up of what exists, prior to making any changes.


gemmaleigh:
This includes the changes from #78 (these styles will be updated as part of the fixed-width PR).

PR #86, opened 2015-07-07
[WIP] Propose gutter multiples for fixed widths

gemmaleigh:
** Please note: These changes aren’t ready to merge and are here to get feedback on how this should work ** 

These changes also include the form reorganisation changes from #85. Ignore those as they will be removed once #85 has been merged.

[Take a look at this commit](https://github.com/alphagov/govuk_elements/commit/e4544a0938264f6e8c49fa3ad60e6ef49c492b3f) which takes @edwardhorsford's suggestion to used fixed widths based off multiples of the gutter.

Not yet considered in this PR:
- These will need to be declared in media queries for tablet and up.
- How to remove the existing 50% width for tablet and up without causing a breaking change

### DIFF
--- a/public/sass/elements/_forms.scss
+++ b/public/sass/elements/_forms.scss
@@ -149,6 +149,43 @@ legend {
 // 6. Form controls
 // ==========================================================================
 
+// NEW form control widths
+// These make use of the gutter multiplier function
+// to give multiples of either 30px ($gutter) or 15px ($gutter-half)
+
+@function gutter-multiplier($gutter, $multiplier) {
+  @return $gutter * $multiplier
+}
+
+// Usage:
+// .my-form-control-width {
+//    width: gutter-multiplier($gutter, 2);
+// }
+
+// TODO: Decide on naming convention for gutter-based widths
+.form-control-1 {
+  width: gutter-multiplier($gutter, 1);
+}
+
+.form-control-2 {
+  width: gutter-multiplier($gutter, 2);
+}
+
+.form-control-4 {
+  width: gutter-multiplier($gutter, 4);
+}
+
+.form-control-8 {
+  width: gutter-multiplier($gutter, 8);
+}
+
+.form-control-16 {
+  width: gutter-multiplier($gutter, 16);
+}
+
+// OLD form control percentage widths
+// Use the fixed widths above in future
+
 // By default, form controls are 50% width for desktop,
 // and 100% width for mobile
 .form-control {


edwardhorsford:
Thanks @gemmaleigh these look good. I may give them a try in anger since 'multiple of gutter' was just a punt. But I think they should be useful and if they work I'll move passports over to this style.

We could add some suggested widths in elements I guess...?

gemmaleigh:
Hi,

I've thought about this a bit more. 
It looks like overkill and it's not clear from the classnames how the two are related. 

From the example page I'm building - we usually have three widths.

.form-control-large - 480px
.form-control-medium - 320px
.form-control-small - 130px

For date of birth - we use smaller widths, 40px and 80px.

We could define additional styles for inline form controls with smaller widths,
or use a multiplier here - based on roughly the number of characters expected 
- but to ensure they are proportion to the other form widths.

edwardhorsford:
Using names like small / medium / large is nicer. I still suggest multiples of something so we have some consistency.

I've just given these a go:
* xx-small: 40px // slightly larger to accommodate 2 characters.
* x-small: 60px
* small: 120px
* medium: 180px
* large: 240px
* x-large: 360px
* xx-large: 480px
* full: full width

The widths go up somewhat exponentially, though there are still quite a few of them - perhaps not all needed. Could get rid of large or x-large.

What they look like as applied to some elements:
govuk-form-widths-test-0 1

gemmaleigh:
Ok, great. 

I'm working on @timpaul's revised form styles at the moment, here's a preview:

forms-styles

I'd like to update the current .form-control { width: 50% } to be 480px.

First name, Last name, Address - are all 480px.

This will also be the default width for textareas.

This then won't be a breaking change or require anyone to add an extra class.
If it's the default, that removes the need for xx-large.

So then we would have:

xx-small (40px)
 x-small  (60px)
    small  (120px)
medium (180px)
     large (360px)

  x-large (set as the default for .form-control)

        full (100%)

The only differences between the sizes you suggest and Tim's are:

Postcode: 
Proposed design:  Postcode (130px)
This would be:       small (120px)

Date of Birth: 
Proposed design:  Day (45px)         Month (80px)   Year (80px)
This would be:       xx-small (40px)  x-small(60px)   x-small(60px)

I think it would be sensible to increase Month and Year to 80px (so make x-small 80px, rather than 60px), 
so the form fields won't need extra spacing between them to line up with the labels above.

Town or City:
Proposed design:  Town (320px)
This would be:       x-large (360px)

These would be larger at 360px.

Let me know what you think.

edwardhorsford:
@gemmaleigh what do you mean by "so the form fields won't need extra spacing between them to line up"?

fyi - based on my research with visually impaired users some weeks back, I'd separately like to suggest each of these fields is moved farther apart.

gemmaleigh:
@edwardhorsford I mean, so it won't look like this:

date-of-birth

Where the label "Month" is wider than it's input, below.

I thought you might try out a native date picker for touch-screen devices, since the issues you observed were all with visually impaired users using VoiceOver on an iPad - that way you would only require the user to activate a single control to choose a date.

edwardhorsford:
Personally I think that's actually better. The boxes are sized to the size of content we expect - I don't think the right edge needs to line up with the label. @timpaul ?

markhurrell:
I like Gemma's version better, looks tidier

gemmaleigh:
@markhurrell pointed out they line up if we use them in this way and looking again at @timpaul's design - I think that's what he had in mind. 

xx-small (40px)
x-small   (80px)
small      (130px) = 40 + 10 + 80
medium  (180px) = 40 + 10 + 130
large       (230px) = 40 + 10 + 180
x-large    (320px) = 80 + 10 + 230 
xx-large  (set as the default for .form-control - 480px)
full (100%)

This way combinations of inline form fields will line up, with a 10px right gutter.

joelanman:
Sorry to put my two cents in really late.

I know 'month' overhanging looks odd, but I'd be worried about a wide month field confusing people as to what to enter, which is more important.

Two small fields and a longer one is clearly 2 digit 2 digit 4 digit, and it's what was researched on Register to vote.

edwardhorsford:
Any progress on this? Would like to be able to point a service at it.

I'm with @joelanman - I consider it more important that the elements are sized to their content than that they line up.

timpaul:
This is partly about whether we want to allow numbers *and* words in the month field. I'm thinking of people (like myself) who sometimes have to count on their fingers to remember whether July is 6 or 7. Any evidence for or against this? I know from Carer's Allowance that the format of any example influences the format people use.

joelanman:
I think it depends on whats being asked. For memorable dates like Date of birth, I've not seen asking for numbers be a problem. For copying from documents, the field should accept, and be formatted for the format on the document.

Considering this proposal is specifically Date of birth, I think we should go with what's been researched, which is 2 2 4 on Register to vote - unless we have research from elsewhere to say otherwise?

cjforms:
It's not just Tim who has to count on his fingers for months. I have mild
dyscalculia and numbers-for-months is painful for me at all times, simply
impossible at times of stress.

I agree that date of birth is one that I have dealt with so many times that
I can manage it more easily as a number for the month.

People with more severe dyscalculia will struggle every single time,
whereas a computer ought to be able to work it out for them fairly easily,

Furthermore, it's easier to disambiguate month-in-day problems (where
people have missed the labels above the boxes, or have been hijacked by
muscle memory and typed irrespective of the labels) if the month is in
letters.

So my question back to the devs is: is it really that hard to parse the
letter inputs?

cheers
Caroline

On Monday, 27 July 2015, Joe Lanman <notifications@github.com> wrote:

> I think it depends on whats being asked. For memorable dates like Date of
> birth, I've not seen asking for numbers be a problem. For copying from
> documents, the field should accept, and be formatted for the format on the
> document.
>
> Considering this proposal is specifically Date of birth, I think we should
> go with what's been researched, which is 2 2 4 on Register to vote - unless
> we have research from elsewhere to say otherwise?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/alphagov/govuk_elements/pull/86#issuecomment-125148698>
> .
>


-- 


Caroline Jarrett
User researcher / forms specialist
Government Digital Service

Twitter @cjforms
07990 570647

My GDS days are usually Monday to Wednesday.

edwardhorsford:
I'd like to suggest that the question of whether the box should accept month names or letter inputs is separate to this PR. 

We're weren't talking about sizing it larger to accommodate month names, but to line up visually - which runs counter to our guidance that form fields should be sized achording to their content.

joelanman:
+1 to ed, fully agree that inputs should accept numbers and words where appropriate, but this is specifically about the visual sizing of the elements.

I'd be very open to research on larger fields - my main worry is that people are unnecessarily confused about what format they should use, but I could be wrong.

kelliematheson:
@gemmaleigh @timpaul I know my comment doesn't relate to what this thread is about, but I can see in your example the radio buttons and checkboxes look completely different, with larger inputs and without the rectangular grey background hit areas.

Is the style for radio buttons and check boxes changing to this? And what brought on the change, as I thought the previous style was thoroughly tested and it tested well?

joelanman:
@kelliematheson the custom radios and checkboxes are not part of this pull request. We are researching them and they are promising. Don't want to derail the discussion on this pull request, but happy to continue on the [design patterns wiki](https://designpatterns.hackpad.com/Radio-buttons-and-checkboxes-8eBuLm9eRaz)

cjforms:
Hi Kellie

In lots and lots of user research, we observed that many people with low
digital skills had a hard time using conventional radio buttons and check
boxes because they aimed for the small round target (radio button) or small
square target (checkbox).

The larger inputs / rectangular grey background areas were an attempt to
solve this problem by trying to communicate to users that they didn't need
to aim for the small targets, but could instead click anywhere on the label.

Despite this, we continued to observe lots of users aiming for the small
targets. Somewhat surprisingly, this included many people with high digital
skills. Because of Fitts' Law
(the current Wikipedia article on Fitts Law
<https://en.wikipedia.org/wiki/Fitts%27s_law> is pretty good), this has a
small impact on a large number of high digital skills users. Even worse, we
observed that there were still lots of low digital skills users who were
continuing to struggle with the small targets. So they have a major impact
on a (relatively) small number of users, but these are users who are at
their limits when trying to do things online so we need to be especially
careful to help them.

My suggestion, therefore, was to make the targets bigger, and to abandon
the grey background thing. Fitts Law say that bigger targets are easier to
aim for and easier to fit. Giving up on the grey background also gives us a
small, but helpful, improvement in legibility. The label and the adjacent
area is still clickable.

The Verify team tested big targets and found that:
- Users with high digital skills aren't fazed in the least by the big
targets - they don't even notice them.
- Users with low digital skills who think that they need to aim for the
target now have a nice big thing to aim for and it's much easier for them.
- If a user aims for the big target and misses slightly, it doesn't matter
because the label and the adjacent area is still clickable
- If a user knows that the label is clickable, they're still fine
- If a user aims for the label and misses slightly, they're also still fine
because the adjacent background area is still clickable. .

Verify has been using big targets in live for a while.

I also ran a small pilot eye-tracking study to compare the two versions
(grey background / big targets) against each other. I was aiming to work
out a way of doing this sort of testing, so I recruited five GDS
colleagues. Each of them did a short task, then filled in a form
(umoderated, no think alound), then they went back over the same task with
me and filled in a form with the same questions (repeating and explaining
their previous experience). As it happened, we wanted to test the questions
in the form anyway so it was a handy opportunity.

The participants knew that we were testing the questions in the form; I
didn't mention the difference in visual presentation, but simply asked the
participants to click a different link for the form in their first attempt
and in the second. None of the participants noticed the difference, and
even more interestingly I kept the switch to myself and the observers
didn't notice either. It didn't make any difference which form they filled
in first.

Obviously a small comparison test with colleagues isn't the last word in
definitive testing, but at least it was a bit of direct comparison that
helps to confirm the Verify findings from testing big targets on their own.

Cheers
Caroline





Caroline Jarrett
User researcher / forms specialist
Government Digital Service

Twitter @cjforms
07990 570647

My GDS days are usually Monday to Wednesday.

cjforms:
@joelanman - thanks for clarifying that.



Caroline Jarrett
User researcher / forms specialist
Government Digital Service

Twitter @cjforms
07990 570647

My GDS days are usually Monday to Wednesday.

edwardhorsford:
Do we have any reason not to proceed? I'd like to make use of this.

gemmaleigh:
Closing this for now. 

This PR isn't the place to discuss custom radio buttons or checkboxes - discussion needs to happen over on the Hackpad - i'll create a new PR for moving to fixed widths for form controls.

cjforms:
Thanks, good idea

PR #89, opened 2015-07-16
Change to 'button-start' class name

joelanman:
The current class name is `button-get-started` which can be confusing and hard to remember if you don't have that text on the button.

The example says 'Start now' - I propose just having `button-start` so it's agnostic to what text is on the button.

Continue to support `button-get-started` as an alias for now, so it doesn't break people's code

### DIFF
--- a/public/sass/elements/_buttons.scss
+++ b/public/sass/elements/_buttons.scss
@@ -36,6 +36,7 @@
 }
 
 // Start now buttons
+.button-start,
 .button-get-started {
   @include bold-24;
   background-image: file-url("icons/icon-pointer.png");


gemmaleigh:
This is called `.button-get-started` as GOV.UK uses `.get-started` as a parent wrapper for `.button`.

I don't object to this, I just think it would be good to show why it has this name.

Here's an example with a [Start now button](https://www.gov.uk/carers-allowance-report-change).

joelanman:
Ah, good to know where it came from. I'm guessing at some point it said "Get started" on GOV.UK. As we don't often use that language any more, as far as I can see, the class name seems a bit arbitrary.

gemmaleigh:
You're absolutely right, this makes much more sense. 

If we'd gone with .button-arrow, it wouldn't matter what the words before the arrow are.

Since we only use these as "start now" buttons, you're right 
- it's shorter and easier to remember and type.

PR #90, opened 2015-08-03
Add link back style

gemmaleigh:
Add styles for back links to match those used on Register to Vote and Passports, as requested by @edwardhorsford.

Use generated content to show a left-pointing black triangle in browsers which support pseudo-elements. For IE8 and below, use a fallback image.

For Firefox, improve the arrow shape by using rgba dotted borders to smooth the arrow edges.


joelanman:
Like this?

image

gemmaleigh:
Just like that one.

back-link

Issue #93, opened 2015-08-26
select.form-control element arrows are hidden in webkit


pflannery:
When using `select.form-control` the drop down arrows are hidden in webkit caused by [this line](
https://github.com/alphagov/govuk_elements/blob/847d3d2e1c5f3efc33a492fce2b3b9d4765ef3fa/public/sass/elements/_forms.scss#L169). 

Before I override this in our project I just wanted to find out the reasoning just in case there is something i have misunderstood, or maybe this is a bug here? :)

thanks.

cjforms:
This isn't an answer - hope someone with technical knowledge will help - just a quick comment that we're not very keen on using select controls because so many users struggle with them. But I'm sure that's not the reason for the point you're making.

gemmaleigh:
Hi @pflannery, that's there to hide any default webkit styling for form controls (shadows, rounded corners) but I hadn't considered it would affect select boxes that way. At present, there isn't a select box style - there was a pull request for this a while back, but as cjforms said, we're keen to avoid them where possible. I think for now, the best solution would be to override this for your own project and I'll look into removing it.

cjforms:
Thanks @gemmaleigh

pflannery:
Thanks for the fast responses @cjforms @gemmaleigh 

> just a quick comment that we're not very keen on using select controls because so many users struggle with them

Fair enough, I had previously echo'd this point to the requirement makers (design, PO etc.) but its something they still want to go ahead and use. @cjforms just to confirm, am I right in saying that the struggle occurs with the disabled or is it the wider audience as well?

cjforms:
It's with anyone with lower digitial skills. 

Also, important to note that anyone can have their digital skills reduced by a combination of any of these factors: stress, being in a hurry, being tired or disoriented due to shift work, things about the actual service such as confusing language, using a second or additional language.

On the topic of  'the disabled'. Disability can be temporary: having an illness such as a cold, headache, or broken arm, or progessive, or intermittent with relapsing/recurring problems such as rhuematoid arthiritis or multiple sclerosis.

pflannery:
@cjforms thanks for the detailed feedback. very interesting.

gemmaleigh:
Fixed in #94.

pflannery:
@gemmaleigh thanks!

PR #94, opened 2015-08-26
Form control test page

gemmaleigh:
This PR creates a new test page for form elements and shows the difference when `-webkit-appearance: none` is removed from `.form-control`. 

Screenshots:

Before:

dd-before

After:

dd-after


gemmaleigh:
This fixes #93.

PR #92, opened 2015-08-13
Split GOV.UK elements content onto separate pages

gemmaleigh:
GOV.UK elements is currently a very long page, which makes the content hard to find.

This pull request splits up the content from each section onto a separate page, and introduces a consistent structure for each page.

Each page has a list of contents at the top of the page, with jump links to sub-sections and a set of headings at the bottom of each page, grouping links to examples, discussion and further reading.

[Here's a preview on a test Heroku app](http://govuk-elements-test.herokuapp.com/)

PR #96, opened 2015-09-01
Improve GOV.UK elements navigation by splitting content onto separate pages

gemmaleigh:
GOV.UK elements is currently a very long page, which makes the content hard to find.

This pull request is a tidier version of #92. 

The content from each section is now split onto a separate page.

Each page has a list of contents at the top of the page, with jump links to sub-sections and a links to examples (if they exist for that section) at the bottom of the page.

It also introduces a breadcrumb to get back to the elements index page and to show which section examples sit within.

The snippets page has been removed, as the code snippets now sit underneath each example.

[Here's a preview on a test Heroku app](http://govuk-elements-test.herokuapp.com/)


cjforms:
This is great and I'm excited to see it happen. Thanks very much.
Just a couple of small changes I'd like to request: 
1. Please can you change the heading 'Forms' to 'Forms elements' and make corresponding changes on the relevant page. We had advice about forms in many places; at some point in the future, I'll be suggesting some cross-links but for now, that will make it clearer that this advice is about elements relevant to forms rather than about everything to do with forms.
2. Please add links to the various separate discussion pages on the Hackpad to each of the split up pages.

Thanks
Caroline

tombye:
We should have some JS to redirect all the existing URLs with fragment identifiers in to the new section.

For example, if someone clicks on the existing URL `http://govuk-elements.herokuapp.com/#guide-layout`, we should redirect them to `http://govuk-elements.herokuapp.com/layout`.

Something like:

```
var getNewRouteFor = function (fragment) {
  var baseURL = 'http://govuk-elements.herokuapp.com',
      newRoutes = {
          'guide-layout': '/layout'
        };

  fragment = fragment.splice(1);
  if (typeof newRoutes[fragment] !== 'undefined') {
    return baseURL + newRoutes[fragment];
  }
  return false
};

var fragment = window.location.hash,
    newRoute;

if (fragment === '') { return; }
newRoute = getNewRouteFor(fragment);
if (newRoute) {
  window.location.href = newRoute;
}
```

public/javascripts/redirect.js
```diff
@@ -0,0 +1,74 @@
+// Redirect all the existing URLs with fragment identifiers in to the new section
+(function () {
+  'use strict';
+
+
+  var getNewRouteFor = function (fragment) {
+    var baseURL = 'http://govuk-elements.herokuapp.com';
```
tombye: Can this be changed to the site origin so the redirects will work outside of heroku?

I was thinking something like: `var baseURL = window.location.href.split('#')[0];`

tombye:
This looks great, thanks for doing all this work! I just have one question: do the breadcrumb styles in `views/layout_example.html` have to be in-page?

I have a feeling there's a reason, I just can't see it. Other than that, and my comment above, this gets my :+1:

public/javascripts/redirect.js
joelanman: `window.location.host` ?

tombye: `window.location.host` misses out the protocol. My suggestion is a bit hackier but yeah, using `window.location.protocol + '//' + `window.location.host` is probably more correct.

gemmaleigh:
Hi @tombye, the example pages use the main.css file - which only includes elements. There's an extra stylesheet to style the page furniture - the styles for the example boxes, the layout for colour swatches and other styles, this stylesheet (elements-page.css) also includes the styles for the Service Manual breadcrumb.

I'll move the breadcrumb styles into elements as a partial for now and update them to match those in [Alphagov - Static](https://github.com/alphagov/static/blob/da8aeeaa749093eab30286d7fc9f965533b66f47/app/assets/stylesheets/helpers/_header.scss#L224) - but these should probably become a mixin and sit as part of the front-end toolkit - in the same way as the phase banner.

gemmaleigh:
Hi @tombye, 

I've moved the breadcrumb styles into a partial for now, within elements.

The baseURL for the redirects has been amended to work outside of Heroku.

I've made @cjforms' request and renamed the "Forms" section "Form elements" as this better describes the content and added external links to the Hackpad, where a Hackpad exists to provide more information on a pattern.

public/javascripts/redirect.js
tombye: For the record, my original suggestion was incorrect as it would result in the inclusion of the pathname too. Thanks for the correction @joelanman.

tombye:
This looks good to me now. As far as I'm concerned you can merge this yourself assuming all the other things mentioned are in.

PR #99, opened 2015-09-03
Add attributes to snippets

alextea:
In order for the inputs to function in code, they must have the `name` attribute. I've added this so that people copying the snippets don't have to add any extra code.

For inputs in a series (checkboxes) I've used the `variable[]` naming convention from PHP, but I don't know if that works in Node/Express.

### DIFF
--- a/views/snippets/buttons_button.html
+++ b/views/snippets/buttons_button.html
@@ -1,1 +1,1 @@
-<button class="button" type="submit">Save and continue</button>
+<input class="button" type="submit" value="Save and continue">
--- a/views/snippets/buttons_button.html
+++ b/views/snippets/buttons_button.html
@@ -1,1 +1,1 @@
-<button class="button">Save and continue</button>
+<button class="button" type="submit">Save and continue</button>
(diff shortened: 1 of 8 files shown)


gemmaleigh:
Since `<button>` is buggy in earlier versions of IE (especially if the form has more than one button) and there's no requirement for other content (e.g. icons or decoration inside the element) - it's probably best just to go with `<input type="button" value="Continue">` - [this is what's used in the examples here](http://govuk-elements.herokuapp.com/errors/example-form-validation-single-question-radio). 

I'll update the button snippet.

alextea:
Is this OK to merge?

tombye:
@alextea I had a few suggestions on final tweaks to the names, just to line them up with the form you established with the waste-types snippet (making 'nationality' plural) and one change to indicate the boolean nature of a checkbox.

I put it all in a branch so you can merge it onto here if you agree:

https://github.com/alphagov/govuk_elements/tree/add-names-suggestion

alextea:
@tombye LGTM :+1:

PR #101, opened 2015-09-24
Remove links from below the submit button

timpaul:
I think we should remove these from the example, as we're pretty much decided that the 'Back' link goes at the top of the page. And 'Skip' links aren't very commonly used.


joelanman:
Agreed

gemmaleigh:
Thanks for this @timpaul. LGTM.

PR #102, opened 2015-09-24
Update guidance on page width and line length

gemmaleigh:
Recommend using a grid split in a 2/3, 1/3 page layout 
- to prevent overly long lines of text.

This applies to all content, not just body text. 
Add this guidance to the layout section.

cc. @edwardhorsford, @timpaul.

### DIFF
--- a/views/guide_layout.html
+++ b/views/guide_layout.html
@@ -34,9 +34,17 @@
   </div>
 
   <h3 class="heading-medium" id="page-width">Page width</h3>
-  <p>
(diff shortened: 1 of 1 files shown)


gemmaleigh:
@edwardhorsford This has an id - page-width, so you could use: http://govuk-elements.herokuapp.com/layout/#page-width to link to it. I'm not sure it needs an extra heading.

edwardhorsford:
I like it!

PR #106, opened 2015-10-06
Fix panel spacing

gemmaleigh:
**Increase the border width for indented 'info' panels**

These are used to highlight supporting information on GOV.UK and have a 10px left hand border.

info-panel

The current style for indented panels of text:

typography gov uk elements

Proposed change to match GOV.UK:

typography inset text - gov uk elements

**Fix the spacing of toggled content within forms** 

Before:

radio-before

checkbox-before

After:

form elements after - gov uk elements

form elements after - checkboxes - gov uk elements

Also, increase the width of the indented panel for toggled content in forms to 5px. All units used elsewhere are multiples of 5px.

### DIFF
--- a/public/sass/elements/_details.scss
+++ b/public/sass/elements/_details.scss
@@ -21,20 +21,16 @@ details {
       outline: 3px solid $focus-colour;
     }
   }
-  
+
   // Underline only summary text (not the arrow)
   .summary {
     text-decoration: underline;
   }
-  
+
   // Match fallback arrow spacing with -webkit default
   .arrow {
     margin-right: .35em;
     font-style: normal;
   }
-  
-  // Remove top margin from bordered panel
-  .panel-indent {
-    margin-top: 0;
-  }
+
 }
--- a/public/sass/elements/_panels.scss
+++ b/public/sass/elements/_panels.scss
@@ -8,21 +8,21 @@
 .panel-indent {
   @extend %contain-floats;
   clear: both;
-  border-left: 4px solid $border-colour;
+  border-left: 10px solid $border-colour;
 
-  padding: 10px 0 10px $gutter-half;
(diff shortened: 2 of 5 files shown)


joelanman:
Looks good, but would this affect <details> elements too? I'm not sure it'd work well there as it's supposed to look like the line is 'coming from' the details triangle

edwardhorsford:
This is good. In fact, passports already uses the 10px - I changed this long ago and forgot about it.

edwardhorsford:
@joelanman here's an example of what it looks like on Passports currently. Note - when the user expands the details element, we extend the visual focus to the full area to indicate what's added.

screen shot 2015-10-06 at 12 36 05
screen shot 2015-10-06 at 12 36 12

joelanman:
yeh personally I think that specific usage looks a bit odd. It'd be nice if the line looked more lined up with the triangle. If it was, the thickness might not be an issue.

joelanman:
Actually I guess the alignment is an existing issue and could be tackled in another PR. So yeh, this one looks good to me.

cjforms:
Gemma - I'm not clear on this PR. 

Are you simply increasing the width of the left-hand border thing? 
In which case: I don't think it makes much difference, so it's fine by me.

Or are you putting the entire bordered content into a box, like a boxed warning?
In which case: please don't. Boxed warnings tend to have the opposite effect to that which is intended: they're meant to draw the eye to the content, but they often (usually?) make the boxed content appear irrelevant to the conversation of the page.

edwardhorsford:
@cjforms this PR is for the left border being made wider.

The yellow bordered box is focus styling.

gemmaleigh:
Hi @cjforms yes, this is just to add a variant with a wider left hand border for highlighting information. 

The only other change is to make the spacing consistent where these panels are used - for toggled form content and inside the details element.

PR #109, opened 2015-10-21
Add missing legends for fieldsets

gemmaleigh:
This pull requests adds a `<legend>` element as the first element inside any `<fieldset>` elements which were missing them.

Where the content of the `<legend>` duplicates a heading which already exists on the page, it is visually hidden.

This ensures the GOV.UK elements examples [follow our guidance on fieldsets](http://govuk-elements.herokuapp.com/form-elements/#form-fieldsets) 

> Use fieldsets to group related form controls. The first element inside a fieldset must be a legend 
> element which describes the group.

--
From Léonie Watson:

The fieldset/legend elements are important. The `legend` provides a group label that's programmatically associated with the fields contained within the `fieldset`. When focus reaches the first field inside the `fieldset` (whether by tabbing or using other screen reader shortcuts), the `legend` is automatically announced before the label of the field.
 
An `h1` (or any other unassociated text) on the page isn't sufficient because there is no programmatic association between it and the form fields.
 
Hiding the `legend` offscreen is a perfectly acceptable way to handle the visual duplication problem though. Don't use display:none; to do this though because it'll hide it from screen readers too.

--

The form example page is also removed, as this is to be replaced with a "one thing per page" set of form examples.  

cc. @aduggin, @tombye


timpaul:
Looks great to me. Is it worth @aduggin taking a look too?

aduggin:
This looks good. I have just become aware of a bug in VoiceOver on iOS where legends are not read out.  When we have a single question and the heading duplicates the legend this will not be an issue - but when there is not a heading duplicating the legend (due to there be multiple questions on a page) we do have an issue.  Not sure what to recommend  in this instance - have set up a test page to do some investigation: https://govuk-accessibility.herokuapp.com/tests/radio-button-group

aduggin:
:+1:

timpaul:
Another reason for 1-thing-per-page. 

Would be good to record which decisions are due to bugs and dodgy support and which are just good usability.

PR #111, opened 2015-10-21
Add box styles for transaction end pages

gemmaleigh:
Add styles for `.box-highlight`, a turquoise box used for transaction end pages.

Show an example at the bottom of the "Colour" section.

colour gov uk elements

@timpaul, @rivalee, @joelanman.

### DIFF
--- a/public/sass/elements/_boxes.scss
+++ b/public/sass/elements/_boxes.scss
@@ -1,0 +1,7 @@
+.box-highlighted {
+  margin: 1em 0 1em 0;
+  padding: 2em 0 1em 0;
+  color: $white;
+  background: $turquoise;
+  text-align: center;
+}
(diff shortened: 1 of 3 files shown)


joelanman:
looks like it could work more generally, for example on Bank Holidays:
https://www.gov.uk/bank-holidays

So it might fit better in Typography? It's like inset text?

timpaul:
We've also got 'panels.scss', which just has .panel-indent in it at the moment. Do we need boxes and panels, or just one?

gemmaleigh:
I was thinking that panels may well be extended to include notices - which would be used for both text with a border to the side of it and text with an icon and a border. For now, boxes are used for coloured backgrounds, admittedly it's probably not the best name. 

Are there any other instances of boxes with coloured backgrounds, apart from alerts?

timpaul:
Ste McC has a grey box for payment summary in GOV.UK Pay.

joelanman:
@stephenjoe1 is that because the whole process that the user came to do isn't complete?

PR #113, opened 2015-11-03
Add plain list style

timpaul:
I've added a plain '.list' style and example. This is primarily for lists of links where we want the extra margins but we don't want bullets (for example, lists of related links in the right hand column).

### DIFF
--- a/public/sass/elements-page.scss
+++ b/public/sass/elements-page.scss
@@ -40,7 +40,8 @@
 }
 
 // Lists
-.list-bullet {
+.list-bullet,
+.list-number {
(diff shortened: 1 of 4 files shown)


gemmaleigh:
Hi @timpaul, I've made these changes over in #114. Let me know if you're happy with those. I've made the default margin for lists 15px for smaller screens and 30px for larger screens and removed the elements page override, so the bottom margin is the same everywhere.

gemmaleigh:
Hi @timpaul, these changes are now in https://github.com/alphagov/govuk_elements/pull/135.
Once #135 is merged, I'll close this PR.

PR #115, opened 2015-11-03
Add changelog and version number

gemmaleigh:
Create an initial development release — 0.1.0, and add a changelog.

Update contributing guidelines.


CONTRIBUTING.md
```diff
@@ -14,3 +14,8 @@ Remember to:
 * [write helpful commit messages](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html)
 * [check the GDS style guide](https://www.gov.uk/design-principles/style-guide) for tips on how to write to our standard
 * [check the service manual](https://www.gov.uk/service-manual/design-and-content) for tips on how to create content
+
+## Versioning
+
+We use [semantic versioning](http://semver.org/), and bump the version
+on master only. Please don't submit your own proposed version numbers.
```
dsingleton: I think we [recommend bumping the version on a branch/PR](https://github.com/alphagov/styleguides/blob/master/rubygems.md#versioning) now (the style guide is rubygems specific, but we it makes sense to do the same).

version.txt
```diff
@@ -0,0 +1 @@
+0.1.0
```
dsingleton: To work with the `jenkins.sh` this file should be called `VERSION.txt`, which is what we use elsewhere

dsingleton:
I've `jenkins.sh` locally, and it worked OK.

CI job is setup, merged so we can test that properly.

PR #116, opened 2015-11-03
Add legal text example

gemmaleigh:
Add guidance and example of warning text, using the exclamation icon.

warning text

### DIFF
--- a/public/sass/elements/_elements-typography.scss
+++ b/public/sass/elements/_elements-typography.scss
@@ -274,3 +274,19 @@ hr {
     color: $border-colour;
   }
 }
+
+// Notice, text with an icon to the left
+.notice {
+  @extend %contain-floats;
+  position: relative;
+  .icon {
+    position: absolute;
+    left: 0;
+    top: 50%;
+    margin-top: -17px; // Half the height of the important icon
+  }
+  p {
+    padding-left: (35 + 30)+px;
+    margin-left: -$gutter-half;
+  }
+}
--- a/public/sass/elements/_icons.scss
+++ b/public/sass/elements/_icons.scss
@@ -1,8 +1,10 @@
 // GOV.UK icons
 
 .icon {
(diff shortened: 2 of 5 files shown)


gemmaleigh:
cc. @edwardhorsford.

If the important icon is only used in this context, we should probably rename it to icon-warning and also update the front end toolkit.

edwardhorsford:
Sounds sensible, but I don't know that it's *only* used for warnings.

Might be good to get the views of some content designers, since they're usually the ones who use it currently.

StephenGill:
Good idea. On GOV.UK, the rule is we only the warning callout if there are legal consequences (eg a fine or prison sentence).

StephenGill:
The warning text from the waste carrier start page might be a better example: https://www.gov.uk/waste-carrier-or-broker-registration

%You can be fined up to £5,000 if you don’t register.%

joelanman:
we use it the same way on Register to vote - something about if you're trying to register with fraudulent info you can get fined/go to jail

timpaul:
I agree we should restrict to things that have legal consequences -
otherwise services will use it for every little thing and it'll soon be
meaningless.

joelanman:
I think it's a good starting point, and we can gather reasons why services want to use it for other things

gemmaleigh:
Thanks @edwardhorsford, @joelanman, @StephenGill, @timpaul.

I've changed the example to match the one from waste carriers and have added the example to the icons section as well as the typography section.

legal -typography gov uk elements

edwardhorsford:
Nice one!

gemmaleigh:
legal text - typography gov uk elements

@tombye, this should be ready to merge - if you have chance to take a look.

robinwhittleton:
Isn’t this a textbook example for a `<strong>` tag?

robinwhittleton:
👍

PR #118, opened 2015-11-10
Move govuk frontend toolkit dependencies

gemmaleigh:
The aim of this PR is to put all the Sass import statements in once place, with the Sass variables and mixins from the govuk front end toolkit first.

It also moves any Sass imports within partials to the top level - so the main.scss file shows all files imported - this removes the imports from the bottom of forms, for block labels, date of birth and validation.


gemmaleigh:
@robinwhittleton - if you have chance to cast you eye over this, that would be great, ta.

robinwhittleton:
LGTM. Will merge.

PR #119, opened 2015-11-17
Don't commit dependencies, bring closer to prototype kit

fofr:
Do not commit dependencies. Everything in `/govuk` comes from the govuk_template or govuk_frontend_toolkit and is pulled in automatically using `npm install` and then grunt tasks on start.

Updates to bring elements closer inline to current prototype kit:
* Use a Hogan.js module
* Move template config to `lib`
* Use `govuk_modules` directory rather than `govuk`

Added a Wraith config to test for regressions (domain is set to `localhost`). On all pages there was 0% difference.

(Elements is still quite different to the latest prototype kit, but the kit isn't versioned and there's no changelog, so it's hard to make them match.)

cc @dsingleton @gemmaleigh


gemmaleigh:
Hi @fofr, thanks for these changes.

I've submitted an [issue to the prototype kit](https://github.com/alphagov/govuk_prototype_kit/issues/80), to tag the repo when the version is changed - which will make it easier to see the differences between the two.

I've tested deploying to Heroku with these changes and that works as expected so I am happy to merge.

cc. @dsingleton @robinwhittleton.

dsingleton:
:100:

PR #123, opened 2015-11-19
Remove Firefox "fix" for arrow edges

gemmaleigh:
This was an attempt to smooth the edges of the left pointing triangle
in Firefox, as they appeared jagged.

Unfortunately, this caused the following Sass compilation error in the Rails govuk prototype template: 

Base-level rules cannot contain the
parent-selector-referencing character '&'.

Since this “fix” is no longer required, removing these rules fixes this.

Screenshots of Firefox and Safari, to compare:

firefox-nofix

safari

### DIFF
--- a/public/sass/elements/_elements-typography.scss
+++ b/public/sass/elements/_elements-typography.scss
@@ -215,14 +215,6 @@ p {
     margin-top: -6px;
   }
 
-  // Improve arrow shape in Firefox
-  @-moz-document url-prefix() {
-    &::before {
-      border-top: 5px dotted rgba(255, 0, 0, 0);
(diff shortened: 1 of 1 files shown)


gemmaleigh:
Hi @robinwhittleton, this one is just removing a Firefox-specific hack, which was causing rails-sass compilation to fail.

Spotted by @alextea.

robinwhittleton:
Yeah, Firefox hasn’t needed anything like that for a while. Happy to merge.

fofr:
It's interesting to see where this code has propagated to:
https://github.com/search?q=user%3Aalphagov+url-prefix&type=Code&utf8=%E2%9C%93

PR #124, opened 2015-11-23
Remove button margins

gemmaleigh:
Spacing for form elements is made consistent using the .form-group
clearing wrapper, remove the top and right margins, currently added to all buttons.

Also remove overrides for spacing underneath buttons within example boxes.

cc. @tyom, @robinwhittleton.


robinwhittleton:
LGTM 👍

PR #125, opened 2015-11-23
Add a colour swatch for $govuk-blue

gemmaleigh:
The Sass variable $govuk-blue exists in the frontend toolkit, for
#005ea5.

Add this swatch to the colour palette.

colour gov uk elements

### DIFF
--- a/public/sass/elements-page.scss
+++ b/public/sass/elements-page.scss
@@ -177,7 +177,8 @@ $palette: (
   ("green", $green),
   ("grass-green", $grass-green),
   ("turquoise", $turquoise),
-  ("light-blue", $light-blue)
+  ("light-blue", $light-blue),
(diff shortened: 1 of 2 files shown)


gemmaleigh:
It is worth noting that this doesn't follow the variable naming convention used by the other semantic colour variables, but should still be a useful addition to the colour palette.

cc. @joelanman @edwardhorsford @alextea.

edwardhorsford:
Thanks @gemmaleigh. I agree a semantic name would be better, but that doesn't stop this PR being a helpful addition - particularly considering this is a frequently used colour on GOV.UK.

gemmaleigh:
Hi @robinwhittleton, if you could please take a look - this is good to merge.

robinwhittleton:
👍

PR #126, opened 2015-11-24
Create an example page to show icons in the front end toolkit

gemmaleigh:
- Create an example page to show icons from the govuk front end toolkit
- Add a mixin to generate the common rules for each icon
- Use the mixin for the icons in the front end toolkit
- Generate classes for each of the 1 - 14 steps icons.

This PR shows the icons available in the front end toolkit and creating this new mixin aims to make it easier to update them in future (this mixin should move into the front-end toolkit).

Here's a screenshot of the example page:

example icons icons and images gov uk elements

cc. @fofr, @robinwhittleton.

### DIFF
--- a/public/sass/elements/_icons.scss
+++ b/public/sass/elements/_icons.scss
@@ -40,7 +40,7 @@
 }
 
 .icon-important {
-  @include icon(icon-important, 34, 34);
+  @include icon(icon-important, 35, 35);
 }
 
 .icon-information {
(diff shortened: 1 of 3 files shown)


gemmaleigh:
Closing following a discussion in the front-end meeting. This would be clearer if a base class and modifier class are set for each icon, will close and re-open a new PR with those changes.

PR #129, opened 2015-11-26
Show icons in the govuk frontend toolkit

gemmaleigh:
- Create an example page and link to it from the Icons and images section.

- Create a mixin to generate the rules for each icon, to avoid repetition. 
The mixin requires the icon name, width and height and optional subfolder - for the step icons.
(The subfolder option can be removed once the step icons are rebuilt using only CSS)

- Use a base class and modifier class for each icon

Here is a screenshot of the Icons example page:

govuk frontend toolkkit icons

cc. @fofr, @robinwhittleton.

### DIFF
--- a/public/sass/elements/_icons.scss
+++ b/public/sass/elements/_icons.scss
@@ -42,7 +42,7 @@
 }
 
 .icon-important {
-  @include icon(icon-important, 34, 34);
+  @include icon(icon-important, 35, 35);
 }
 
 .icon-information {
--- a/public/sass/elements/_icons.scss
+++ b/public/sass/elements/_icons.scss
@@ -72,3 +72,31 @@
     @include icon(icon-step-#{$i}, 23, 23, icon-steps);
   }
 }
+
+// Propose replacing step icons with circles
+.circle {
+  display: inline-block;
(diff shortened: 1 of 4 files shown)


gemmaleigh:
Propose a `.circle` class to replace the step icons. 

create css circles to use instead

I'm not sure where the step icons are currently used, I couldn't find them in any of the alpahgov repositories, apart from the prototyping app.

public/sass/elements/_icons.scss
```diff
@@ -1,244 +1,102 @@
...
 }
 
-.icon-step-10 {
-  width: 23px;
-  height: 23px;
-  background-image: file-url("icons/icon-steps/icon-step-10.png");
+// Propose replacing step icons with circles
+.circle {
+  display: inline-block;
+  @include border-radius(100%);
```
robinwhittleton: Can we make this 50%, just in case future specs define something different when border radiuses butt up against each other?

gemmaleigh:
@robinwhittleton, if this could please be merged ahead of the PR to add legal messages
https://github.com/alphagov/govuk_elements/pull/116

PR #131, opened 2015-11-30
Copy base styles from govuk template

gemmaleigh:
For users who want to use govuk_frontend_toolkit and govuk_elements, to make use of GOV.UK design patterns.

govuk_elements has the [govuk_template](https://github.com/alphagov/govuk_template) and [govuk_frontend_toolkit](https://github.com/alphagov/govuk_frontend_toolkit) as dependencies.

To use govuk_elements without the govuk_template, any base styles (which are currently set by the template) are copied to the new partial `_base.scss`.

Ideally, this file would use [Normalize.css](http://necolas.github.io/normalize.css/) as its starting point, which can then be customised to meet the design's requirements.

cc. @fofr, @robinwhittleton.


robinwhittleton:
LGTM

PR #133, opened 2015-12-01
[Do not merge, for discussion]. Don't style html elements without a class or parent class

gemmaleigh:
Closing this PR, as it is split into three new PRs.

1. Require the use of a class to apply any govuk elements styles
https://github.com/alphagov/govuk_elements/pull/134

2. Use a common base class and modifier class pattern for lists, panels and icons.
https://github.com/alphagov/govuk_elements/pull/135

### DIFF
--- a/public/javascripts/vendor/details.polyfill.js
+++ b/public/javascripts/vendor/details.polyfill.js
@@ -133,10 +133,10 @@
         var twisty = document.createElement('i');
 
         if (openAttr === true) {
-          twisty.className = 'arrow arrow-open';
+          twisty.className = 'details-summary-arrow arrow-open';
           twisty.appendChild(document.createTextNode('\u25bc'));
         } else {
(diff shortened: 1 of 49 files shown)


edwardhorsford:
Out of interest, what's the thinking behind not styling element selectors - particularly where they're unlikely to change?

Alternately, would an approach be to have the styles under `.details`, but then style the element with ```details {
   @extend .details;
}```

timpaul:
I like the idea of using classes for styling elements, but then making it
easy for people to extend elements with those styles. Then people can
choose how they want to work.

What about an extra Sass file that does all the extending of the basic
typographic elements, but that's commented out by default? User's could
then choose to activate it and to define a class that it's scoped to.

gemmaleigh:
Hi @timpaul, that sounds good. I was thinking that would sit in the prototype kit.

timpaul:
Yep - sounds like a plan

PR #134, opened 2015-12-04
Require the use of a class to apply any govuk elements styles

gemmaleigh:
Create new classnames for `.hr`, `.fieldset`, `.details`, `.details-summary`.

Require a parent class of `.text` for scoped typography styles. 

These changes mean that the govuk_frontend_toolkit can be imported (as main.scss does) followed by govuk_elements and it will have no effect on the project importing it, until one of its  classes are used.

Rather than apply a 19px base to `<main>`, this has been moved to .`text`, so if you'd like to use a 19px base, then the `.text` wrapper must wrap text blocks. 

Assuming govuk_template isn't being used (its element selectors are duplicated in #131), then the only remaining places where element selectors are used are: `elements/reset.scss`. This was copied from [alphagov/static](https://github.com/alphagov/static/blob/master/app/assets/stylesheets/helpers/_reset.scss) and should be removed (but this is out of the scope of this PR).

cc. @robinwhittleton, @dsingleton - it would be good to get these changes merged first, as they affect most of the template files, then I can submit additional pull requests for the remaining changes in #133.

### DIFF
--- a/public/sass/elements/_forms.scss
+++ b/public/sass/elements/_forms.scss
@@ -13,7 +13,7 @@
 // ==========================================================================
 
 // .fieldset is used to apply styling to fieldsets and to group more than one .form-group
-.fieldset {
+fieldset.fieldset {
   @extend %contain-floats;
   width: 100%;
 }
--- a/public/sass/elements/_forms.scss
+++ b/public/sass/elements/_forms.scss
@@ -3,37 +3,20 @@
 
 // Contents:
 //
-// 1. Helpers
-// 2. Form wrappers
-// 3. Form labels
-// 4. Form hints
-// 5. Form controls
-// 6. Form control widths
+// 1. Form wrappers
+// 2. Form labels
+// 3. Form hints
+// 4. Form controls
+// 5. Form control widths
 
-// 1. Helpers
+// 1. Form wrappers
 // ==========================================================================
 
-// Fieldset is used to group more than one .form-group
-fieldset {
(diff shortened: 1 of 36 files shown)


public/sass/elements/_forms.scss
```diff
@@ -3,38 +3,21 @@
...
+// 5. Form control widths
 
-// 1. Helpers
+// 1. Form wrappers
 // ==========================================================================
 
-// Fieldset is used to group more than one .form-group
-fieldset {
+// .fieldset is used to apply styling to fieldsets and to group more than one .form-group
+.fieldset {
```
robinwhittleton: Are we happy with changes like these? I can see people now writing `<div class="fieldset">` which then breaks accessibility for a bunch of people. Given the overall simplicity of our pages I’m not worried about selector complexity impacting performance so would it be better to have `fieldset.fieldset` as the selector? Anything that enforces accessibility from within our styling is a good idea (imo of course).

gemmaleigh: Absolutely, that's perfectly sensible. We don't want this class applied to anything other than a `<fieldset>`. I'll update it.

tommyp: This is a good suggestion :+1:

gemmaleigh:
Closing this PR in favour of renaming `_base.scss` to `_govuk-template-base.scss` and moving any HTML element styles set by govuk elements into `_base.scss`.

PR #135, opened 2015-12-04
Use a common naming convention

gemmaleigh:
Use a common base class and modifier class pattern for lists, panels, icons (the changes to icons have already been made in #129).

Also use a common naming convention for grid columns, add `.column-one-third` and `.column-one-half`, update the examples to show use of these new classnames (while .column-third and .column-half remain, this isn't a breaking change - but aims to make our classnames for layout easier to guess).

cc. @robinwhittleton.


robinwhittleton:
👍

PR #139, opened 2015-12-08
Make the list of files govuk elements imports a partial

gemmaleigh:
This PR moves the files that govuk elements imports from main.scss to a new file, govuk-elements.scss.

It aims to make it easier to keep govuk-elements up-to-date in the govuk prototype kit.

The advantage of using a partial here is that it doesn't generate an extra stylesheet, so users of the prototype kit can choose to either have the default setup (importing all files listed in govuk-elements.scss) or to pick and choose files to import from those listed, by copying the list of files back into main.scss.

There are also new groupings for the imports for the front end toolkit - to show which files contain Sass mixins, functions or variables.

The phase banner styles are also moved from the elements example page, to a new partial.

cc. @robinwhittleton, @timpaul, @aduggin, @tombye.


robinwhittleton:
Like it

PR #140, opened 2015-12-08
Fix missing list classes

gemmaleigh:
Add missing `.list` class to lists.

`list-style-type: none`, should probably be a default, so the `.list` class isn't needed to apply this.

This relates to #134, these HTML element styles should be set in `_base.scss`.
I'll raise an issue.

### DIFF
--- a/public/sass/elements-page.scss
+++ b/public/sass/elements-page.scss
@@ -88,6 +88,12 @@
     color: $hm-government;
   }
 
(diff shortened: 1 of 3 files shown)


gemmaleigh:
This adds missing class attributes which should have been applied in: https://github.com/alphagov/govuk_elements/commit/a8b31029339a6135ee8240661acb3c32b2218733

PR #142, opened 2015-12-09
Update the package name to use the govuk convention

gemmaleigh:
Both the govuk template and govuk frontend toolkit use underscores,
rather than hyphens in the name. Use an underscore here for consistency.

Also update the description.

Once this has been merged, the [govuk prototype kit PR](https://github.com/alphagov/govuk_prototype_kit/pull/103) to keep govuk elements up-to-date in the kit can be updated too.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
-  "name": "govuk-elements",
-  "description": "GOVUK prototyping app in Express",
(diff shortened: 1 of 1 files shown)


gemmaleigh:
Leaving this for now as it'll break things for anyone who is using the existing name.

PR #152, opened 2016-01-05
Use locally installed grunt

gemmaleigh:
Reference the locally-installed version of Grunt for the 'test' task to check that the app runs.

This updates the govuk elements Travis file to match the [govuk prototype kit](https://github.com/alphagov/govuk_prototype_kit/pull/112).


quis:
:+1:

PR #156, opened 2016-01-06
Update package json to publish govuk-elements-sass

### LINKED ISSUE #155: Make govuk_elements SASS it's own module
It'd be really neat to be able to install just the SASS from govuk elements much in the same way as the govuk_frontend_toolkit

So probably using npm we could have something like:

```bash
npm install --save govuk_elements_sass
```

We are currently copying it and other people are installing _all_ of elements just to get at the SASS so making this into it's own module would solve lots of problems.

gemmaleigh:
One solution, to resolve #155 - would be to create a govuk_elements_sass_npm repo [as already exists for the govuk frontend toolkit](https://github.com/alphagov/govuk_frontend_toolkit_npm), with only the Sass files whitelisted, as below.

Here, updating `package.json` would make it possible to request only the Sass files, without the rest of the govuk elements application.

These changes make the files available within `public/sass`, excluding those in `.npmignore`.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,9 @@
+public/sass/main-ie6.scss
+public/sass/main-ie7.scss
+public/sass/main-ie8.scss
+public/sass/main.scss
+public/sass/elements-page-ie6.scss
+public/sass/elements-page-ie7.scss
+public/sass/elements-page-ie8.scss
+public/sass/elements-page.scss
+public/sass/prism.scss
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - 0.1
+  - 4.0
 before_install:
   - npm install -g grunt-cli
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,17 +13,26 @@ notifications:
   email: false
 
 deploy:
+  matrix:
   # Deploy master to staging (govuk-elements-test.herokuapp.com)
   - provider: heroku
     api_key:
-        secure: MSMzniJpxcjP1jbsjqaAR/bU/FtGRrYYrrIekzaXNXtUiwNOsnd2MMMiBiuwzRaqHlEZzUkh70jyCuAoztnMzr5vAlwy9IW5BnGhacbDZsyFZ8AS1Z0Glq0l7q0nh1akCEVYhdGyABrc2NMaAqtaVovEbAfZYCrSX8Uu0CRbPfY=
+      secure: MSMzniJpxcjP1jbsjqaAR/bU/FtGRrYYrrIekzaXNXtUiwNOsnd2MMMiBiuwzRaqHlEZzUkh70jyCuAoztnMzr5vAlwy9IW5BnGhacbDZsyFZ8AS1Z0Glq0l7q0nh1akCEVYhdGyABrc2NMaAqtaVovEbAfZYCrSX8Uu0CRbPfY=
     app: govuk-elements-test
     on: master
   # Deploy tagged build to production (govuk-elements.herokuapp.com)
   - provider: heroku
     api_key:
-        secure: MSMzniJpxcjP1jbsjqaAR/bU/FtGRrYYrrIekzaXNXtUiwNOsnd2MMMiBiuwzRaqHlEZzUkh70jyCuAoztnMzr5vAlwy9IW5BnGhacbDZsyFZ8AS1Z0Glq0l7q0nh1akCEVYhdGyABrc2NMaAqtaVovEbAfZYCrSX8Uu0CRbPfY=
+      secure: MSMzniJpxcjP1jbsjqaAR/bU/FtGRrYYrrIekzaXNXtUiwNOsnd2MMMiBiuwzRaqHlEZzUkh70jyCuAoztnMzr5vAlwy9IW5BnGhacbDZsyFZ8AS1Z0Glq0l7q0nh1akCEVYhdGyABrc2NMaAqtaVovEbAfZYCrSX8Uu0CRbPfY=
     app: govuk-elements
+    on:
+      all_branches: true
+      tags: true
+  # Release tagged commits to npm
+  - provider: npm
+    api_key:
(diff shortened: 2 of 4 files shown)


glenjamin:
This would be great!

I think the package.json would also need to trim down the dependencies so we only pull in the parts that the sass depends on, possibly only leaving the frontend toolkit as a dep.

I think @jonhurrell would be keen to see this also.

gemmaleigh:
It can't happen in this repo - so I haven't removed the dependencies here. 

We'd need to create a new repo, with a `package.json` file like this one, with only the govuk frontend toolkit listed as a dependency.

Thoughts? @tombye, @dsingleton, @robinwhittleton, @fofr. 

This would also make it [easier to keep the Sass files for the prototype kit up to date](https://github.com/alphagov/govuk_prototype_kit/commit/9a9c5e089468931eff4638154208ac78b9e09f22).

If this is something you'd like to use in your project can you please leave a comment here to let me know if this solution would work for you. It would be good to find out who else is keen to make use of this.

jonhurrell:
Any dependency management gets a 👍from me. However, I'm not convinced npm is the best tool for the job for sass partials; given it's a package manager specific to the node.js back-end environment, whereas Bower is a little more agnostic.

quis:
@gemmaleigh NPM has `dependencies` and `devDependencies`. They let you specify which packages get installed when you’ve developing elements, and which get installed when you’re pulling it in as a dependency. [Good Stackoverflow answer with more detail](http://stackoverflow.com/a/22004559/147318).

easternbloc:
My 2 pence is strongly :-1: for bower. There is no need for yet another package manager. Plenty of people distribute front end codez via npm.

joelanman:
+1 for NPM

https://www.quora.com/Why-use-Bower-when-there-is-npm

tombye:
I'm also :+1: for not using bower. If you're compiling Sass then you're probably using [node-sass](https://github.com/sass/node-sass), whether it's wrapped in a Grunt, Gulp or other build tool plugin or not, so your project will have a `package.json`. That being the case, you might as well use it.

We should be looking at it in terms of what we need from our dependancy management tool and whether or not the chosen solution can provide it. NPM seems to allow us to do everything we need to and is already on projects with front-end builds.

As an extra note, it's worth pointing to [the argument 18f make for not using bower](https://pages.18f.gov/frontend/recommendations/dependency-management/) too which goes into the more technical reasons.

easternbloc:
@gemmaleigh I think this is a really nice solution. Obvs the big thing here is to be very descriptive about what you're getting if you install it via npm. Other than that :ok_hand:

ztolley:
Looks good

gavboulton:
This is looking good.

On passports we're using Elements in a couple of different ways. We're using the Sass in https://github.com/UKHomeOffice/passports-frontend-toolkit/tree/master/assets/stylesheets/govuk-elements (which we're manually copying, hence why @easternbloc raised the issue that triggered this PR). The current version we're using is quite old so maintenance is a problem, especially as passports-frontend-toolkit is now being widely used across the Home Office. We're also using the details element polyfill in a private repo, which is also manually copied.

What about exposing the entire public directory?

gavboulton:
As a small note, you have "govuk_frontend_toolkit" in dependencies and devDependencies.

jonhurrell:
@easternbloc You're right plenty of people do it, doesn't mean to say it's the right approach though.
@joelanman Plenty of replies in the posted thread that side on using bower for front end assets.

Either way just happy they'll be abstracted out.

gemmaleigh:
Hi @gavboulton, thanks - I'll remove the duplicate.

I'd rather not expose everything in /public/ - as most of the JavaScript already exists in the front end toolkit and [it would be better to use the module pattern documented there](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/docs/javascript.md).

There's a plan to move the details summary polyfill to the frontend toolkit too.

That just leaves the images, the majority of which are only needed for the govuk elements examples, everything else is already in the front-end toolkit (aside from the breadcrumb separator image, which I've made a note to move).

gavboulton:
@gemmaleigh, great. It sounds like you've got everything covered.

We're using Browserify for our client-side JS dependency management so would blend the moved details polyfill in using this.

gemmaleigh:
Things left to do for this PR:
- [x] Add a licence
- [ ] Ensure all images used by the Sass files exist in the govuk frontend toolkit's images folder
- [ ] Remove "private: true" and publish to npm

Then, this fixes #155.

Also, raise issues as reminders to:
- [ ] Move the details summary polyfill to the front end toolkit

easternbloc:
@gemmaleigh and documentation to explain what happens when you install it!

easternbloc:
@gemmaleigh ah you just added that! Nice :+1:

gemmaleigh:
@easternbloc let me know if you think the [readme is missing anything](https://github.com/alphagov/govuk_elements/commit/806f5d52857c2b3a81bd61c1e2298d5213a6792f#diff-04c6e90faac2675aa89e2176d2eec7d8R93). Cheers!

gavboulton:
:+1:

gemmaleigh:
#### Other things to do before this can be merged.

This PR ensures that only the Sass files are published to npm. The image assets used by the Sass files must be moved to the front end toolkit. These images are the separator images for the breadcrumb and the fallback arrow image for the back link.

- [ ] Move the [breadcrumb Sass partial and images to the front-end toolkit](https://github.com/alphagov/govuk_frontend_toolkit/pull/236)
- [ ] Bump the version for the govuk frontend toolkit
- [ ] Update the version of the gov frontend toolkit used by govuk elements
- [ ] Import the breadcrumb Sass from the updated govuk front end toolkit and replace the elements breadcrumb partial
- [ ] Move the fallback image for the "Back" link to the front end toolkit
- [ ] Bump the version for govuk elements

Then, this fixes #155.

glenjamin:
We've just pulled in your work-in-progess to try it out by depending on the git commit directly:

```sh
npm install --save alphagov/govuk_elements#806f5d52857c2b3a81bd61c1e2298d5213a6792f
```

The files have been pulled in correctly, and everything looks good.

The only minor issue is the `engine` field, which is generating a warning that could be removed.

```
npm WARN engine govuk-elements-sass@1.0.0: wanted: {"node":"0.10.x"} (current: {"node":"5.3.0","npm":"3.3.12"})
```

tombye:
Looks good to me :+1:

easternbloc:
:ok_hand:

gavboulton:
:+1:

PR #157, opened 2016-01-13
Bump the frontend toolkit to 4.6.0

gemmaleigh:
Update the frontend toolkit to 4.6.0

- use the new breadcrumbs partial from the front end toolkit
- use the same markup as the govuk breadcrumb component
- update the classname used by the breadcrumbs
- add the ARIA role="breadcrumbs"


fofr:
:+1: Good to merge when green

PR #167, opened 2016-01-28
Make error messages plural rather than using (s)

edwardhorsford:
We avoid using `(s)` in content on GOV.UK - the preference is to use the plural form where it's ambiguous.

I know these are just examples, but would be good to match content style.


gemmaleigh:
Thanks for these @edwardhorsford. :+1:

PR #168, opened 2016-01-28
Fix the breadcrumbs for Icons and images

gemmaleigh:
Change the section to `icons-images`, so the middle breadcrumb links back to the “Icons and images” section.


mcgoooo:
:shipit:

PR #171, opened 2016-01-29
Fix warning for npm engine

gemmaleigh:
When `govuk-elements-sass` is installed, a warning is shown for the engine -
setting this to be greater than or equal to 4.0.0 fixes this.

@edwardhorsford, let me know if this fixes #165.

PR #172, opened 2016-01-29
Bump the version to 1.1.2

gemmaleigh:
# 1.1.2

- Add missing 2/3 width grid wrappers to example pages (#173)
- Update max for date pattern to 2016 (#175)
- Fix warning for the npm engine field (#171)
- Bump the govuk frontend toolkit to 4.6.1, update the $link-hover-colour swatch. (#170)
- Remove 'files' from package.json and amend .npmignore (#169)
- Fix the breadcrumbs for the 'Icons and images' section (#168)
- Make error messages plural rather than using (s) (#167)

cc. @tombye, @robinwhittleton, @fofr.


robinwhittleton:
Happy with this.

PR #173, opened 2016-02-05
Add missing 2/3 width grid wrappers to example pages

gemmaleigh:
All main content should sit in a 2/3 column grid.
Add `.grid-row` and `.column-two-thirds` divs where they are missing.

This is most noticeable on the Errors and validation pages - the error message summary box should also be constrained by a wrapper which is 2/3 of the page width.

Screenshot before:

before - example form validation errors gov uk elements

Screenshots after:

example form validation errors gov uk elements

2 example form validation errors gov uk elements

### DIFF
--- a/views/examples/example_date.html
+++ b/views/examples/example_date.html
@@ -7,13 +7,18 @@
 
   {{>breadcrumb}}
 
-  <h1 class="heading-xlarge">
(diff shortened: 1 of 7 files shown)


robinwhittleton:
LGTM

PR #175, opened 2016-02-08
Update max for date pattern to include this year

### LINKED ISSUE #174: using cursor keys, you can't get 2015 or 2016 in the year input
http://govuk-elements.herokuapp.com/form-elements/example-date/

gemmaleigh:
This should be `2016`, and fixes #174.

cc. @robinwhittleton, @joelanman, @henryhadlow.

### DIFF
--- a/views/snippets/form_date.html
+++ b/views/snippets/form_date.html
@@ -21,7 +21,7 @@
 
       <div class="form-group form-group-year">
         <label for="dob-year">Year</label>
-        <input class="form-control" id="dob-year" name="dob-year" type="number" pattern="[0-9]*" min="0" max="2014">
+        <input class="form-control" id="dob-year" name="dob-year" type="number" pattern="[0-9]*" min="0" max="2016">
       </div>
     </div>
 


joelanman:
does this highlight its a fragile thing to even use? Services would have to remember to update every year - seems likely that it'll sometimes be broken.

henryhadlow:
I was thinking the same thing. Could we just set it to longtime in the
future and then have additional validation post-submission?

Also, it should be min="1899"

https://en.m.wikipedia.org/wiki/List_of_the_verified_oldest_people

On 8 February 2016 at 13:16, Joe Lanman <notifications@github.com> wrote:

> does this highlight its a fragile thing to even use? Services would have
> to remember to update every year - seems likely that it'll sometimes be
> broken.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/alphagov/govuk_elements/pull/175#issuecomment-181362708>
> .
>



-- 
Henry Hadlow
Product Manager
GOV.UK Notify
<https://gds.blog.gov.uk/2015/10/05/status-tracking-making-it-easy-to-keep-users-informed/>
• petition.parliament.uk
Government Digital Service <https://gds.blog.gov.uk/>

robinwhittleton:
Services would presumably make it dynamic (e.g. `<%= Time.now.year %>` in Rails). As for oldest people, I can imagine some sort of historical lookup service. I randomly worked on a [Australian convict search](http://www.lincolnshire.gov.uk/convictssearch.aspx) system for local government a while back and while this doesn’t have birth date specifically there’s no reason why it couldn’t have.

joelanman:
what's the user need for a max date?

PR #176, opened 2016-02-16
Copy base styles set by govuk template

gemmaleigh:
This pull request removes any GOV.UK template-specific variables and mixins.
It aims to ensure that the govuk-elements-sass package can be used without govuk_template (so only the govuk_frontend_toolkit is a dependency).

The base partial is renamed to govuk-template-base.scss to make it clear where these styles have come from.

Variables used only by the GOV.UK template for header and footer colours, eg. $footer-background are removed. 

The `media-down(mobile)` mixin is swapped with `media(tablet)` for external links, [to match the template](https://github.com/alphagov/govuk_template/commit/86106bfb332ff5c0346b92310a4117daeb2d8c8d).

cc. @dsingleton, @robinwhittleton.

### DIFF
--- a/public/sass/_govuk-elements.scss
+++ b/public/sass/_govuk-elements.scss
@@ -32,7 +32,7 @@
 // Base (unclassed HTML elements)
 // These are predefined by govuk_template
 // If you're not using govuk_template, uncomment the line below.
-// @import "elements/base";                       // HTML elements, set by the GOV.UK template
+@import "elements/govuk-template-base";           // Base styles set by GOV.UK template
 
 // Objects (unstyled design patterns)
 @import "elements/layout";                        // Main content container. Grid layout - rows and column widths
--- a/public/sass/elements/_govuk-template-base.scss
+++ b/public/sass/elements/_govuk-template-base.scss
@@ -1,7 +1,22 @@
 // https://github.com/alphagov/govuk_template/blob/master/source/assets/stylesheets/_basic.scss
 
 // basic styles for HTML5 and other elements
-html, body, div, h1, h2, h3, h4, h5, h6, article, aside, footer, header, hgroup, nav, section {
+html,
+body,
+div,
+h1,
+h2,
+h3,
+h4,
+h5,
+h6,
+article,
+aside,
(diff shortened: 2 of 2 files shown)


mcgoooo:
@gemmaleigh amazing, looks good!

gemmaleigh:
It would be good to fix this.

cc. @robinwhittleton, @tombye, @dsingleton, @fofr.

This exists because of the [GOV.UK template](https://github.com/alphagov/govuk_template/pull/201), overrides the $grey colours from the frontend toolkit. 

It would also be good to remove `media-down`, [where it still exists](https://github.com/alphagov/govuk_template/search?utf8=%E2%9C%93&q=media-down).

dsingleton:
Hi, i'm not sure i'm in a good position to review this - I don't have the context behind the change :(

It's not clear to me why this file exists and duplicates the styles from `govuk_template`. I'm also a confused why the diff is adding a new file that duplicates [`_base.scss`](https://github.com/alphagov/govuk_elements/blob/fix-base-styles-set-by-govuk-template/public/sass/elements/_base.scss) - was that supposed to be moved/removed?

gemmaleigh:
For anyone using GOV.UK elements, but not using GOV.UK template - this file (which has been renamed to show where its contents originated from) - copies the base styles set by the GOV.UK template.

@fofr had this issue, as did @mcgoooo. 

These GOV.UK template base styles use an old `$grey` variable, which doesn't exist in the frontend toolkit and also use the media-down mixin.

e4592d7 copies the $grey variable that is missing from the frontend toolkit

127d829 swaps the order for external link sizes, so [media-down](https://github.com/alphagov/govuk_template/blob/8fa253ff88dd3bb43aba813f0cb7b6da269995b0/source/assets/stylesheets/styleguide/_conditionals2.scss) isn't being used.

Ideally, these base styles would exist in the govuk frontend toolkit, to avoid duplication here.

robinwhittleton:
Let’s talk this over in person tomorrow then make a decision.

robinwhittleton:
All looks good.

PR #178, opened 2016-02-18
add margin-bottom to .code blocks

joelanman:
This has no effect on inline elements with the `code` class, but importantly adds a margin-bottom to any block elements.

### DIFF
--- a/public/sass/elements/_elements-typography.scss
+++ b/public/sass/elements/_elements-typography.scss
@@ -251,6 +251,7 @@ p {
   background-color: $highlight-colour;
   border: 1px solid $border-colour;
   padding: 4px 4px 2px 4px;
+  margin-bottom: $gutter;
 }
 
 // Horizontal rule style


gemmaleigh:
Hi @joelanman, it would be good to know where you intend to use this. 

For formatted blocks of code, GOV.UK elements uses `<pre>` as a wrapper, which sets the bottom margin. 

These styles are defined in prism.css.

```
pre[class*="language-"] {
  padding: 1em;
  margin-bottom: $gutter;
  overflow: auto;
}
```

Would it work for you if these styles were copied to typography.scss? (I'm assuming you want to include code examples in the prototype kit).

joelanman:
oops I forgot to reply here - yeh code examples in the kit is what I'm trying to achieve, so yeh whatever Element's using would be good. Then again - is Elements the place to put this? It's not useful on GOV.UK as far as I know. Might be best just defined in Elements and the Kit separately.

gemmaleigh:
@joelanman are you happy to close this in favour of adding prism.js to the prototype kit and setting any styling for `<pre>` in prism.css?

joelanman:
sure thats a better idea, thanks!

PR #180, opened 2016-02-19
[For discussion, do not merge] Aim to make margins and spacing more consistent

gemmaleigh:
I think this can be split into 2 PRs.

1. To set consistent margins for lists, panels and tables to match body text:
15px (`$gutter-half`) for smaller viewports, 20px (`$gutter-two-thirds`) for larger viewports 

2. To make the margins on `.heading-size` classes in GOV.UK elements more sensible.

For large headings - make the margins consistent with the govspeak
titles - for both 48px and 36px titles, use margins of 15px for
smaller viewports and 45px (`$gutter * 1.5`) for larger viewports.

For smaller headings, use 15px for
smaller viewports and 30px (`$gutter`) for larger viewports. 

### Before:

before - example margins and spacing typography gov uk elements

### After: 

after - example margins and spacing typography gov uk elements

### 
With a background colour to make the margins easier to see.

### Before:

before - pink - example margins and spacing typography gov uk elements

### After: 
after - pink - example margins and spacing typography gov uk elements

### DIFF
--- a/public/sass/elements/_elements-typography.scss
+++ b/public/sass/elements/_elements-typography.scss
@@ -67,12 +67,12 @@ main {
 .heading-xlarge {
   @include bold-48();
 
-  margin-top: em(15, 32);
-  margin-bottom: em(30, 32);
+  margin-top: $gutter-half;
+  margin-bottom: $gutter-half;
 
   @include media(tablet) {
-    margin-top: em(30, 48);
-    margin-bottom: em(60, 48);
+    margin-top: $gutter * 1.5;
+    margin-bottom: $gutter * 1.5;
   }
 
   .heading-secondary {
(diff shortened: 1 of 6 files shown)


edwardhorsford:
IMHO these are really large margins for headings. When a 36px heading is used on one question per page or body copy ends up really far down. They look good for longform copy, but I'm not so sure for services.

fofr:
When looking at the margins for govuk_components I also found that one size doesn't fit all. It'd be nice to do some separate investigation into what works for different types of content, and then define that. cc @markhurrell

gemmaleigh:
Hi @edwardhorsford I'll add an example service page so we can compare. Here the heading spacing has been reduced, which will ensure body copy sits further up the page.

gemmaleigh:
This PR needs a a set of "One question per page" examples to be built to compare the heading sizing and spacing on form pages, vs. longer text pages.

I'm closing this for now, as I'm away until the 15th of April. I'll reopen once I'm back.

PR #182, opened 2016-02-23
Remove invalid ARIA role

gemmaleigh:
This was copied from the [govuk-breadcrumb component](https://github.com/alphagov/static/pull/737), which has since removed `role=“breadcrumbs”`.

cc. @robinwhittleton, @fofr.

### DIFF
--- a/views/includes/breadcrumb.html
+++ b/views/includes/breadcrumb.html
@@ -1,5 +1,5 @@
 <div class="breadcrumbs">
-  <ol role="breadcrumbs">
+  <ol>
     <li><a href="{{site.baseurl}}/">GOV.UK elements</a></li>
     {{#section}}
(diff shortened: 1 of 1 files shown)


robinwhittleton:
There’s an open question as the whether this should be replaced by a different role in the [static issue](https://github.com/alphagov/static/pull/737). Can we hold off on this until it’s resolved? (cc @aduggin)

gemmaleigh:
@robinwhittleton, I'll mark as not ready to merge - I created this PR as a reminder to update based on what gets merged into the component.

fofr:
Whether the pattern is improved or not, the role is invalid and should be removed. I don't see the ongoing pattern discussions as a blocker to this PR or the one on static.

gemmaleigh:
@fofr, I agree. I was waiting on a decision on the PR on static, but I would prefer to get this merged and improve the pattern later.

Issue #121, opened 2015-11-18
Make default table text size the same as body copy


timpaul:
We should assume that table text is the same 19px size as standard body copy, unless specified otherwise. Would be nice if the .font-xsmall class could be applied to an entire table though.

gemmaleigh:
Following the discussion in the front end meeting. we'll keep the text size for tabular data at 14px for smaller screens, 16px for tablet and up and for special cases, for example - the "Check your answers" pattern, the text size will be increased to the standard body copy size of 19px.

edwardhorsford:
@gemmaleigh I missed the frontend meeting. Could you summarise the discussion / the points for and against?

gemmaleigh:
@timpaul @edwardhorsford, [here's the PR to increase the font size for tables to 19px](https://github.com/alphagov/govuk_elements/pull/185), so tabular data is the same size as body copy. 

It also adds a new `.table-font-xsmall` class (as Tim suggested) to recreate GOV.UK style smaller tables.

gemmaleigh:
#185 fixes this.

PR #185, opened 2016-02-24
Use 19px as the default font size for tabular data

gemmaleigh:
All body copy should be the same, at 19px, this includes tabular data.

GOV.UK has tables set at 16px, these are to stand out from large areas
of body copy at 19px.

Add a .table-font-xsmall class, to recreate these smaller table styles.

After:

16px
16px - data gov uk elements

19px
19px - data gov uk elements

### DIFF
--- a/public/sass/elements/_tables.scss
+++ b/public/sass/elements/_tables.scss
@@ -8,8 +8,8 @@ table {
 
   th,
   td {
-    @include core-16;
-    padding: em(12, 16) em(20, 16) em(9, 16) 0;
+    @include core-19;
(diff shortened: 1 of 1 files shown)


gemmaleigh:
This fixes #121.

gemmaleigh:
cc. @timpaul, @edwardhorsford.

edwardhorsford:
Not tested, but I support the idea in principle!

dsingleton:
:+1: from me. It might be worth adding a table with a lot of columns, or wide column content, to illustrate the `table-font-xsmall`

timpaul:
:+1:

PR #187, opened 2016-02-25
Use aria-describedby to associate the hint with the form fields

gemmaleigh:
This will cause screen readers to announce the hint (after they
announce the legend and/or form label for the field).

cc. @aduggin.


joelanman:
nice!

robinwhittleton:
👍

PR #188, opened 2016-03-02
Remove $white background color on selected option

richardTowers:
Some projects (e.g. Verify) use different default colours.
Having a $white background for selected options causes
unacceptably low contrast (in our case white on white).

See alphagov/verify-frontend#18

### DIFF
--- a/public/sass/elements/forms/_form-block-labels.scss
+++ b/public/sass/elements/forms/_form-block-labels.scss
@@ -57,7 +57,6 @@
 
 // Add selected state
 .js-enabled label.selected {
-  background: $white;
   border-color: $black;
 }
 


richardTowers:
@gemmaleigh - PR as requested.

If it helps determine why this was here in the first place, it looks like it arrived in 3d1a899e6b5e.

gemmaleigh:
Hi @richardTowers, thanks for this. 

These `.selected` styles should *only* apply to block labels, which have a grey background `$panel-colour`, and that's why the white background is set here for the selected state. 

I appreciate this style is affecting - all - labels, not just block labels (which is further reaching than it should be), I'll amend so that the selected styles only apply to block labels.

I'm not sure this will fix the issue you have on Verify, as there isn't a modifier for this style of labels (it assumes that block-labels are grey on a white background).

I'm wondering if still it makes sense to use the .block-label class, afaik Verify uses custom radio buttons and checkboxes which don't look like .block-labels at all.

cc. @robinwhittleton, thoughts?

robinwhittleton:
In the end I fixed this in alphagov/verify-frontend directly. This probably shouldn’t be removed here.

Issue #166, opened 2016-01-26
Radio buttons jump up and down on Safari when activated


robinwhittleton:
Radio buttons, when activated drop down by about 10 pixels in Safari 8 and 9 on OS X. They jump back up to their original positions when the animation ends.

([filed the original issue in the wrong place](https://github.com/alphagov/govuk_frontend_toolkit/issues/221))

gemmaleigh:
Removing the width and height for the input fixes this (as in Safari, the width and height don't increase the size of the form control).

gemmaleigh:
#296 fixes this.

PanosParalakis:
Removing the width or height for the input fixes this but  the position is changed. Position had to be updated as well. I have made a pull request which is fixing this #323 .

robinwhittleton:
Fixed by https://github.com/alphagov/govuk_elements/pull/296 (and released in elements 2.2.0)

PR #194, opened 2016-03-14
[Discuss:] GOV.UK Verify style radios and checkboxes

gemmaleigh:
Larger hit area radio buttons and checkboxes.

Screenshot before:

before - example radio buttons and checkboxes form elements gov uk elements

Screenshot after:

example radio buttons and checkboxes form elements gov uk elements

Pros:
- a larger hit area for the control
- the selected radio or checkbox is more obvious (background, border and bold text)
- This fixes issue: #166 (jump in Safari when radio button is selected)

Cons:
- This requires extra markup within each block label `<span><span class="inner"></span></span>`


edwardhorsford:
I'm *strongly* in favour of bigger radio and checkboxes. They're much better.

I'm not sure we still need the selection state / boxes with these changes. Though they are good on mobile.

Other thoughts:
* The checkboxes and radios are not vertically centred
* We should avoid bolding up the selected text. This can cause jarring movement, particularly if it causes text to wrap to two lines.

edwardhorsford:
NB - if we want to be hacky we could use `-webkit-text-stroke-width: 0.5px;` or similar which will give a bit of a bold appearance, but not change the width of the text.

edwardhorsford:
Any examples of hover styles? For reference, I coincidentally made this last week:
http://address-lookup-prototype.herokuapp.com/option-2/delivery-address (search for anything).

tombye:
The description doesn't seem to explain the pros of using this design for radio buttons/checkboxes (over just making the native control bigger).

I'd be interested to know the reasons for the change to the windows 8/10/mobile style.

aduggin:
We'll need to check these for accessibility - need to make sure they work in assistive technologies as expected.
Gemma - could you add test cases to the sandbox?
I've added example for the current style for testing:
https://govuk-accessibility-sandbox.herokuapp.com/patterns/checkboxes-group/
https://govuk-accessibility-sandbox.herokuapp.com/patterns/radio-buttons/

edwardhorsford:
I've seen some services with a fair amount of text within the radios, including bold for headings. Would probably be good if we avoid changing the weight on selection for this reason too.

tombye:
This may be possible without the extra markup (using some CSS3 features), if it we were ok with a certain level of coverage for the CSS (above version 8 for IE and possibly some other older clients).

I've done a test for what I mean, be interested in some reckons on it (needs proper testing):

https://github.com/tombye/custom-field-controls

timpaul:
Thrilled to see this discussion happening ;-)

I second Ed's comment about the grey backgrounds - research has suggested they don't really get people to click the labels, so they're really just reducing contrast. Mobile might be a different case.

Can we design an A/B test to see if what effect these changes have?

joelanman:
I'm in favour of big, clear controls for all browsers. Even when browsers do support making built-in controls bigger, theyre still often not as clear as they could be, or fit the GOV.UK style.

In research so far, I don't think the borders or grey boxes add much in terms of usability, and they detract in terms of contrast, and extra visual noise. So we don't have them on Verify, however, we haven't done loads of research on mobile.

gemmaleigh:
I'm closing this for now, as it's not in a mergable state and it was raised for discussion only. 
I'm away until the 15th of April. I'll reopen once I'm back and we can look at a better solution then.

PR #183, opened 2016-02-24
Use GOV.UK modules js

gemmaleigh:
- Copy JS from the govuk_frontend_toolkit
- Use the GOV.UK modules JS
- Find and auto-start any modules
- Create new modules for toggling hidden content when a radio button or checkbox is selected
- Update examples which show toggle functionality

PR #198, opened 2016-03-21
Fix toggled content

### LINKED ISSUE #197: Toggled content not displaying
When toggling content on the Elements app, the toggled content doesn't display:
govuk-elements-toggle-content-bug

Observed on mac on Chrome 49

gemmaleigh:
Rather than setting `display:none` and `display:block` using `show()`; and
`hide()`, add and remove the `.js-hidden` class.

[I think this was introduced by this change to the GOV.UK template](https://github.com/alphagov/govuk_template/commit/0b6684331ba0aac717dfe4dbe3a2743e2108e54f), which sets both display:none and visibility:hidden, and the visibility:hidden isn't removed when show and hide are used.

This fixes #197.

I've also removed .js-enabled .js-hidden from GOV.UK elements as this is defined in the template.

It is worth noting that #183 has a much better implementation of the JS for toggles.

cc. @robinwhittleton @edwardhorsford.

### DIFF
--- a/public/javascripts/application.js
+++ b/public/javascripts/application.js
@@ -34,14 +34,14 @@ function ShowHideContent() {
             var $groupDataTarget = $('#' + groupDataTarget);
 
             // Hide toggled content
-            $groupDataTarget.hide();
+            $groupDataTarget.addClass('js-hidden');
             // Set aria-expanded and aria-hidden for hidden content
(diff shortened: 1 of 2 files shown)


robinwhittleton:
Happy with this, will merge.

PR #209, opened 2016-04-20
Add an example using Gulp to compile the govuk-elements-sass files

gemmaleigh:
Explain how to install the govuk-elements-sass package.

Explain how to use Gulp's `includePaths` setting to set the path to the files imported from the
govuk_frontend_toolkit and the govuk-elements-sass package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,13 +33,15 @@ Ignore the `/public/sass/elements-page.scss` files, these exist to style the pag
 
 ## Using the govuk-elements-sass package
 
-If you would like to import the govuk elements Sass files into your project, you can do so using:
+To install the govuk-elements-sass package, use:
 
     `npm install govuk-elements-sass`
 
-This will install the files within `public/sass` to `node_modules/govuk-elements-sass`.
+This installs the package as well as the packages it depends on to the local `/node_modules/` folder.
 
-The GOV.UK frontend toolkit dependencies are listed at the top of the `_govuk_elements.scss` partial, shown below:
+The GOV.UK frontend toolkit scss dependencies must be imported first, as these contain Sass variables, mixins and functions.
+
+The `_govuk_elements.scss` partial shows which files are required.
 
     // GOV.UK frontend toolkit
     // Sass variables, mixins and functions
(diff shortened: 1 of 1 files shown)


gemmaleigh:
cc.@quis, @robinwhittleton, @tombye. 

If anyone has a moment to check this makes sense, I'd be most grateful.

README.md
```diff
@@ -37,9 +37,10 @@ If you would like to import the govuk elements Sass files into your project, you
 
     `npm install govuk-elements-sass`
 
-This will install the files within `public/sass` to `node_modules/govuk-elements-sass`.
+This will install the files within `public/sass` to `node_modules/govuk-elements-sass` and the govuk_frontend_toolkit to
```
tombye: This is total reckon but this read a bit odd to me because the fact the package is an NPM package seems implied.

It makes more sense as a concept to me to say something like: "The `govuk-elements-sass` files will be installed into your `node_modules` folder as `node_modules/govuk-elements-sass`. The files included in this package are restricted so only those in the `public/sass` folder will be copied across.

This is just my opinion though and I'm far from the typical user so be interested to hear what others think of this.

README.md
```diff
@@ -64,14 +65,58 @@ The GOV.UK frontend toolkit dependencies are listed at the top of the `_govuk_el
     // Functions
     // @import "url-helpers";                         // Function to output image-url, or prefixed path (Rails and Compass only)
 
-
-Either copy these files your Sass directory, or configure the `includeFiles` path if you’re using a task runner like Grunt.
+Either copy these files your Sass directory, or configure `includePaths` if you’re using a task runner like Gulp.
```
tombye: I think `copy these files your Sass directory` is missing an `into`.

gemmaleigh:
Hi @tombye, I've had a go at addressing your comment, thanks - they were really helpful.

README.md
```diff
@@ -65,13 +67,62 @@ The GOV.UK frontend toolkit dependencies are listed at the top of the `_govuk_el
     // @import "url-helpers";                         // Function to output image-url, or prefixed path (Rails and Compass only)
 
 
-Either copy these files your Sass directory, or configure the `includeFiles` path if you’re using a task runner like Grunt.
+Add the `node_modules` directory to the `includePaths` property of your Sass plugin - if you're using a task runner like Gulp, to reference the location of these files.
```
quis: Should say `node_modules/govuk-elements-sass` directory to match the example below I think.

gemmaleigh: Good spot @quis, I've added both as it needs to know the paths to `node_modules/govuk_frontend_toolkit` and `node_modules/govuk-elements-sass`.

PR #213, opened 2016-04-26
Update the link to departmental colours

gemmaleigh:
The organisation colours have been split into a new file, update the
link to this partial.

cc. @fofr


robinwhittleton:
Good catch

Issue #200, opened 2016-03-21
Highlight box colour contrast too low for WCAG AA


maxf:
Highlight boxes (as visible at the bottom of https://govuk-elements.herokuapp.com/colour/ ) have a too low colour contrast (3.17) to comply with WCAG2 level AA.
Offending line is:
https://github.com/alphagov/govuk_elements/blob/master/public/sass/elements/_components.scss#L7
Suggest: #0c857b as the nearest colour that satisfies the requirement

sanjaypoyzer:
Isn't this only an issue if the colour is used directly on a white background? Is there an example (apart from the swatch) where this is done?

For the swatch, we should probably be putting a border around the element as with the white swatch.

gemmaleigh:
Two examples spring to mind:

1. The GOV.UK Bank Holidays page
https://www.gov.uk/bank-holidays

2. The "Confirmation page" pattern
http://govuk-prototype-kit.herokuapp.com/examples/confirmation-page

Both use `#28a197`, or `$turquoise`.

Both examples use white text on the turquoise background.

Since it passes AAA for large text and the smallest font size for the text on this colour on the Bank Holiday page is 18px (for smaller viewports) and 24px (for wider viewports), I'd be keen to keep the background colour the same and add guidance to ensure that text used on this background colour isn't ever smaller than 18px or 14px bold.

--

Foreground: #FFFFFF - Background: #28A197

The contrast ratio is: 3.17:1

Text failed at level AA
Text failed at level AAA

**Large text passed at level AA**
Large text failed at level AAA

1.4.3 Contrast (Minimum):  Text (and images of text) have a contrast ratio of at least 4.5:1, except if the text is pure decoration.  Larger scale text (at least 18 point or 14 point bold) or images of text can have a contrast ratio of 3:1. (Level AA)

1.4.6 Contrast (Enhanced): Text (and images of text) have a contrast ratio of at least 7:1, except if the text is pure decoration.  Larger scale text (at least 18 point or 14 point bold) or images of text can have a contrast ratio of 4.5:1. (Level AAA)

Note: Fonts that are extraordinarily thin or decorative are harder to read at lower contrast levels.

sanjaypoyzer:
Ah, sorry I misunderstood - I thought we were talking about `$highlight-colour`.

In the case of the turquoise boxes, I agree that if it passes AAA with large text, we should make that clear in the guidance.

edwardhorsford:
[...]

PR #214, opened 2016-04-27
Add guidance on using turquoise backgrounds

sanjaypoyzer:
Fixes the problem discussed in issue #200.

### DIFF
--- a/views/guide_colour.html
+++ b/views/guide_colour.html
@@ -379,6 +379,10 @@
 
   <h3 class="heading-medium" id="examples">Examples</h3>
   <h4 class="heading-small">Transaction end page</h4>
+  
+  <p>
+    Note that the white text on turquoise background only passes accessibility standards if the font size is at least 18px, or 14px bold. 
+  </p>
 
   <div class="example">
 
--- a/views/guide_colour.html
+++ b/views/guide_colour.html
@@ -381,7 +381,7 @@
   <h4 class="heading-small">Transaction end page</h4>
   
   <p>
-    Note that the white text on turquoise background only passes accessibility standards if the font size is at least 18px, or 14px bold. 
+    Note that the white text on turquoise background has a contrast ratio of 3:1, so only meets the WCAG2 Level AA accessibility standard when large text is used. Make sure you use at least font size 18px, or 14px bold when using the highlight box.
   </p>
 
   <div class="example">


views/guide_colour.html
```diff
@@ -379,6 +379,10 @@ <h4 class="heading-small">GOV.UK blue</h4>
 
   <h3 class="heading-medium" id="examples">Examples</h3>
   <h4 class="heading-small">Transaction end page</h4>
+  
+  <p>
+    Note that the white text on turquoise background only passes accessibility standards if the font size is at least 18px, or 14px bold. 
+  </p>
```
gemmaleigh: 'accessibility standards' is a little vague, could this be revised to "only meets [
[WCAG2 Level AA](https://www.w3.org/TR/WCAG20/#visual-audio-contrast-contrast), with a contrast ratio of 3:1 for large text - when the font size is at least 18px, or 14px bold.

sanjaypoyzer:
@gemmaleigh Not quite sure why the CI failed there...

gemmaleigh:
I'd like to revise this to:

Make sure text on a coloured background is legible.

To meet <a href="https://www.w3.org/TR/WCAG20/#visual-audio-contrast-contrast">WCAG2 Level AA</a>, you need to make sure the:

- font size is at least 18px, or 14px bold
- contrast ratio is 3:1

--

I've run this past the content designers on the Service Manual team.

@sanjaypoyzer let me know if you're happy to amend and squash those two commits, otherwise I can.

edwardhorsford:
@gemmaleigh Does that mean that essentially services shouldn't use `core-19` in those boxes? (since the font size goes down to 16px on mobile).

sanjaypoyzer:
Yeah sure, that looks clearer. Will just wait for the clarification on the `core-19` point before doing it.

gemmaleigh:
Yes, the example as-is would fail WCAG2 Level AA for smaller viewports, where `.core-19` for body text at 19px decreases to 16px.

If we don't intend to change the colour of the background - then the font size can either be increased, or the 19px white text on a turquoise background could use `.bold-19` which would pass.

sanjaypoyzer:
Should we add that to the guidance as well, or do you think it's clear enough to be inferred?

gemmaleigh:
This was added in #481, closing as this has now been resolved.

Issue #201, opened 2016-03-24
Disabled button styles do not have sufficient contrast


edwardhorsford:
The style for disabled buttons does not have sufficient contrast - it's 2.48:1.

![screen shot 2016-03-24 at 10 36 23](https://cloud.githubusercontent.com/assets/2204224/14014770/896b0148-f1ac-11e5-85c3-23367a903e1e.png)

IMHO in most cases we should avoid using disabled styles, but where they are needed, we should come up with a style that can be seen easily.

joelanman:
Seen easily and obviously disabled seems like conflicting goals.
Can we just ban disabled buttons? If a button action can't be done, message the user about why and how to rectify when they press it.

matthewford:
Disabled buttons are useful to prevent double form submissions

joelanman:
In that use case, is fading the button the best visual solution? Maybe a spinner or something?

matthewford:
Although you could so something fancy with animations.
![stripe](https://cloud.githubusercontent.com/assets/215/14019004/85244c18-f1c8-11e5-9895-e9b985918319.gif)

Fading is quite a common pattern for disabled buttons - even this github comment from does it.

edwardhorsford:
I think we should separate the discussion of whether 'disabled' styles should suggest from what they look like.

For the purposes of this issue, it would be good if we had a style that supports users with visual impairments whilst also looking correct.

matthewford:
I agree a better disabled style that is accessible is a good thing. I use the disabled styles, so wouldn't want to see them go.

edwardhorsford:
Interestingly it seems meeting contrast checks for inactive elements is *not* a W3C accessibility requirement.

"Incidental: Text or images of text that are part of an inactive user interface component, that are pure decoration, that are not visible to anyone, or that are part of a picture that contains significant other visual content, have no contrast requirement."

Link: https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html

See also these interesting threads:
http://ux.stackexchange.com/questions/80881/do-disabled-buttons-still-need-to-be-contrast-compliant-for-accessibility
http://ux.stackexchange.com/questions/60359/accessible-disabled-state

matthewford:
That thread also argues that disabled buttons are not 'pure decoration'

sanjaypoyzer:
Have we got any research that shows people with visual impairments have a worse experience because of this issue?

[...]

PR #215, opened 2016-04-27
add guidance for disabled buttons

sanjaypoyzer:
Clarification for the usage issues outlined in #201 (although not the core issue of contrast).

### DIFF
--- a/views/guide_buttons.html
+++ b/views/guide_buttons.html
@@ -82,7 +82,7 @@
 
   <h3 class="heading-medium" id="button-disabled">Disabled buttons</h3>
   <p>
-    Disabled buttons should be set at 50% opacity.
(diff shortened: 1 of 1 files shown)


gemmaleigh:
I think we should open this one up for discussion.

I'd like to say something along these lines:

--

Don’t disable buttons, unless there’s a good reason to.

** Describe suitable reason to disable buttons here **

If you have to disable buttons, make sure they look like this...

--

Any thoughts on a bettter example? 

cc. @edwardhorsford, @joelanman, @timpaul.

gemmaleigh:
I'm summarising from #201:

If you have to disable buttons, make sure they look like you can't click them. 

An example of a useful disabled option is a calendar with greyed out days where no bookings are available.

Provide another way for the user to continue, add an error message or text to explain why the button is disabled.

gemmaleigh:
Closing in favour of #304.

PR #223, opened 2016-05-12
Don't copy the govuk frontend toolkit images to an icons subfolder

gemmaleigh:
Remove the icons subfolder.

These assets are now copied to `/public/images`, so the folder structure for
these assets is the same as that of the [govuk frontend toolkit's images folder](https://github.com/alphagov/govuk_frontend_toolkit/tree/master/images).

For the govuk-elements-rails gem, which uses the govuk_frontend_toolkit gem, the path to the image assets will be the same.

cc. @robinwhittleton @robmckinnon.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -64,7 +64,7 @@ module.exports = function (grunt) {
         expand: true,
         src: '**',
         cwd: 'node_modules/govuk_frontend_toolkit/images/',
-        dest: 'govuk_modules/public/images/icons/'
+        dest: 'govuk_modules/public/images/'
       },
 
     },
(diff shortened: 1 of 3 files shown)


gemmaleigh:
I'm going to merge this, as it'll fix the issue with the current broken external links on Heroku. 

There's a further issue, which is that govuk elements isn't importing the [URL helpers from the toolkit](https://github.com/alphagov/govuk_elements/blob/master/public/sass/_govuk-elements.scss#L23), but is [setting a fallback path](https://github.com/alphagov/govuk_elements/blob/master/public/sass/elements/_helpers.scss#L6) to `public/images/` which may be interfering with anyone using the URL helpers.

PR #225, opened 2016-05-16
Add `tabindex-1` to fix the skiplink in Safari

gemmaleigh:
The GOV.UK template sets a skiplink to `#content` (just tab on any GOV.UK page and hit ENTER when "Skip to main content" is highlighted).

This allows users to navigate to the main content area using only the keyboard. Pressing enter when focus is on the skiplink should then set focus on the #content area.

In Safari, there's a bug where this doesn't happen unless `tabindex="-1"` is set on the destination element (in this case `<main id="content">`).
Here's the bug report: https://bugs.chromium.org/p/chromium/issues/detail?id=37721 

This pull request adds:

- `tabindex="-1"` to the `<main>` element wherever it appears in a template

cc. @aduggin.

### DIFF
--- a/public/sass/elements-page.scss
+++ b/public/sass/elements-page.scss
@@ -4,6 +4,13 @@
 // Elements page styles
 // ==========================================================================
 
+// Pressing enter when focus is on the skiplink sets focus on the #content area
+// Remove the blue outline
+
+#content {
+  outline: none;
+}
+
 // These are example styles, used only for the elements pages
 
 .elements-index {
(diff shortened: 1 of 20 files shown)


fofr:
Can this bug still be reproduced? What versions of Safari are affected and what's their usage?
See problems here: https://github.com/alphagov/govuk_template/pull/321

fofr:
On GOV.UK which does not have tabindex set:
Testing on Safari 10.1.1 it works with VoiceOver on desktop.
Testing on iOS 10.2.1 it works in mobile Safari with VoiceOver on.

Issue #221, opened 2016-05-09
Hidden text (progressive disclosure) example no longer has blue triangle in Firefox Nightly / Developer Edition


gunjam:
Hi,

If you use [Firefox Developer Edition](https://www.mozilla.org/en-GB/firefox/developer/) or [Firefox Nightly](https://nightly.mozilla.org/) the blue triangle no longer renders for progressive disclosure exampes using the `<summary>` and `<details>` tags:

![No triangle!](http://i.imgur.com/VqvO3KC.png)

I imagine this is due to native support for `<summary>` and `<details>` [recently landing in Firefox dev builds](http://www.otsukare.info/2016/04/19/summary-details). I believe it'll arrive in release Firefox in version 49 (mid September).

I don't know if this is a CSS, JS polyfil or Firefox issue, but there is a [ticket on bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=591737) tracking `<summary>` and `<details>` support if more info is required.

Thanks!

robinwhittleton:
It’s a Firefox issue: by default they’re styling the new `summary` element with `display: list-item` to get a marker. We apply a style of `display: inline-block` to get the focus outline to wrap the text. I filed a bug a few days ago with Mozilla (https://bugzilla.mozilla.org/show_bug.cgi?id=1270163) but the temporary fix for us will be to remove the `display: inline-block` styling; focus state wrapping is less of an issue than the missing arrow.

gunjam:
Awesome, thanks.

gemmaleigh:
I think a nice solution to this would be move the [fix to revert Firefox's display back to list-item](https://github.com/alphagov/govuk_elements/pull/227) to a [shame.scss file](http://csswizardry.com/2013/04/shame-css/). 

It makes it clear that it's code that we're not happy about and it can then be included by anyone who would like to fix this in Firefox Developer Edition or Firefox Nightly.

robinwhittleton:
Done 😊

PR #227, opened 2016-05-19
Fix the summary arrow in recent Firefox

robinwhittleton:
This is a pretty hacky fix so I’ve tried to describe it as accurately as possible in the comments. It can be removed if either of the following happens:

* Mozilla fixes their summary arrow with `display: inline-block`
* Mozilla remove their support for `@-moz-document`

Fixes https://github.com/alphagov/govuk_elements/issues/221


robinwhittleton:
What does everyone else think. Worth adding?

robinwhittleton:
Hang on, that breaks in older versions of Firefox. Will revise.

gemmaleigh:
I've tested this in Firefox Developer Edition and it works nicely, screenshots below.

What is currently in master:

screen shot 2016-05-20 at 11 32 29

This branch:

screen shot 2016-05-20 at 11 34 00

thanks @robinwhittleton 👏

PR #229, opened 2016-05-20
Consistent spacing underneath block labels and toggled content

### LINKED ISSUE #222: Form label and hint inside a fieldset
I think some improvements need to be made for form labels and hints within a fieldset for example a YES/NO question.

In the examples presented. All checkboxes, radio buttons have effectively a medium title which is all very nice for questions that follow the radio or checkbox structure. (Single page questions)

The problem I have found is when you work with standard form inputs and introduce radios etc... you need consistency with the form patterns for everything as an example:

http://onlineplea.herokuapp.com/#/your-details

This page demonstrates consistency, however I have had to add a hack which I am not proud of. Perhaps time can be spent looking into this issue:

**CSS**
```
.form-hint {
  margin-bottom: 0;
  padding-bottom: em(7);  
}

legend span ~ span {
  margin-top: em(-7);
} 
```

**HTML**

```
<legend>
   <span class="form-label-bold">Label</span>
   <span class="form-hint">Help</span>
</legend>
```

This fixes spacing issues. But I don't like it.

Perhaps I have missed something, or approached this the wrong way. Any advice and or direction would be good.

### LINKED ISSUE #224: Inline radio improvements (spacing)
Inline radio buttons have a margin bottom of 15px. This margin is for spacing, if conditionally revealing content exists. Very nice.

The problem is when conditionally revealing content does not exist. Where we simply want a Yes/No answer. In this case it makes no sense to have a margin bottom value. And as a result, this puts the space consistency out of rhythm.

My suggestion would be to apply the margin top on the panel if it sits within an inline field set.

This would then fix the spacing issue for inline radio when no conditional content is required

gemmaleigh:
> This PR sets consistent spacing between form labels and controls (1.), removes additional spacing between block labels - the large hit area radio buttons and checkboxes (2.) and adds more examples to the Form elements example page (3).

------------

1. Set 10px spacing underneath form labels, form hint text and an error message, when it is shown (see screenshot for 3).

2. Remove the spacing underneath block labels when they are inline (screenshots below).

There is currently a 10px bottom margin, to set spacing between block labels.

before - opened

The above screenshots shows extra spacing underneath the YES/NO question.
This bottom margin works well when they are stacked, but when they are inline - it should be removed.

There is also extra bottom spacing underneath the toggled content (seen under National Insurance number and Country Name fields).

after

This fixes #224.

3. This is the updated Form elements example page:

This fixes  #222, #224.

### DIFF
--- a/public/sass/elements/_forms.scss
+++ b/public/sass/elements/_forms.scss
@@ -46,7 +46,7 @@ textarea {
   margin-bottom: $gutter;
 
   @include media(tablet) {
-    margin-bottom: $gutter*2;
+    margin-bottom: $gutter * 2;
   }
 }
 
--- a/public/sass/elements/_forms.scss
+++ b/public/sass/elements/_forms.scss
@@ -35,6 +35,21 @@ textarea {
 // 2. Form wrappers
 // ==========================================================================
 
+// Form section is used to create spacing between form sections
+.form-section {
+  @extend %contain-floats;
+  @include box-sizing(border-box);
+
+  float: left;
+  width: 100%;
+
+  margin-bottom: $gutter;
+
+  @include media(tablet) {
+    margin-bottom: $gutter*2;
+  }
+}
+
 // Form group is used to wrap label and input pairs
 .form-group {
   @extend %contain-floats;
--- a/public/sass/elements/_forms.scss
+++ b/public/sass/elements/_forms.scss
@@ -35,7 +35,7 @@ textarea {
 // 2. Form wrappers
 // ==========================================================================
 
-// Form section is used to create spacing between form sections
+// Form section is used to wrap .form-group and has twice its margin
 .form-section {
   @extend %contain-floats;
   @include box-sizing(border-box);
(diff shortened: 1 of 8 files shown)


trevorsaint:
Hi Gemma I will take a look at this. Big thanks for looking into it :)

trevorsaint:
Hi Gemma.

That seems to have fixed one part of it. There is another that still exists that you may or may not be aware of. Please see example below:

screenshot-localhost-3000-2016-05-20-14-12-35

To get around this issue I had to do the following:

```
*.form-hint {
  margin-bottom: 0;
  padding-bottom: em(7);  
}
legend span ~ span {
  margin-top: em(-7); // Far from ideal!
}
```
Not nice. But vertical spacing is fixed as a result. I do not wish to modify core GOV element files.

Any chance of looking into this one?

Big thanks 

Trevor

public/sass/elements/_forms.scss
```diff
@@ -35,6 +35,21 @@ textarea {
 // 2. Form wrappers
 // ==========================================================================
 
+// Form section is used to create spacing between form sections
```
robinwhittleton: Happy to merge when this comment is updated.

PR #232, opened 2016-05-20
Use a branch as an alias of the latest release

dsingleton:
This is to make diffing master against  whatever the latest released version
of the package is. This allows us to easily see what has been committed to
master but not yet released in a consistent way, eg with a diff like:

https://github.com/alphagov/govuk_elements/compare/latest-release...master

This means we don't need to know the latest release is tag `v4.10.0` (for example)

This also lets us track un-released package changes in tools like the deploy-lag-radiator
(https://github.com/dsingleton/deploy-lag-radiator)


gemmaleigh:
@dsingleton thanks, this is really useful! 💯

PR #234, opened 2016-05-23
Fix path for rails and node environments

gemmaleigh:
Check to see if the `image-url` function exists

If `image-url` is not defined (if we are not in a Rails environment), set
a path to `/public/images`.

This will fix the issue where elements is overriding `$path` for the [govuk_elements_rails gem](https://github.com/alphagov/govuk_elements_rails). 

cc. @robinwhittleton @dsingleton.


robinwhittleton:
Checked this with a custom `govuk_elements_rails` build and it works fine.

PR #235, opened 2016-05-23
Bump the govuk frontend toolkit to 4.12.0

gemmaleigh:
This will bring in the updated button padding, will remove the
@font-face declaration for Helvetica, and remove the @viewport
declarations (they’re added back in to govuk_template in 0.17.2).


gemmaleigh:
cc. @robinwhittleton

PR #238, opened 2016-05-30
Correct guidance: numbered lists should be written as full sentences

StephenGill:
What we currently say about lists isn't quite right: numbered lists are written as full sentences. See the style guide: https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#bullet-points-steps

... and, for example, this page: https://www.gov.uk/guardians-allowance/how-to-claim

Apologies for any errors in the html.

@gemmaleigh

### DIFF
--- a/views/guide_typography.html
+++ b/views/guide_typography.html
@@ -134,7 +134,7 @@
 
   <h3 class="heading-medium" id="typography-lists">Lists</h3>
   <p class="text">
-    List items start with a lowercase letter and have no full stop at the end.
+    Bulleted list items start with a lowercase letter and have no full stop at the end. Numbered list items are written as full sentences, so they start with an uppercase letter and have a full stop at the end.
   </p>
 
 <div class="example">
--- a/views/snippets/typography_lists.html
+++ b/views/snippets/typography_lists.html
@@ -1,15 +1,15 @@
+<p>Here is a bulleted list:</p>
 <ul class="list list-bullet">
-  <li>here is a bulleted list</li>
   <li>vivamus sagittis lacus vel augue laoreet rutrum faucibus dolor auctor</li>
   <li>vestibulum id ligula porta felis euismod semper</li>
   <li>integer posuere erat a ante venenatis dapibus posuere velit aliquet</li>
 </ul>
 
+<p>Here is a numbered list.</p>
 <ol class="list list-number">
-  <li>here is a numbered list</li>
-  <li>vivamus sagittis lacus vel augue laoreet rutrum faucibus dolor auctor</li>
-  <li>vestibulum id ligula porta felis euismod semper</li>
-  <li>integer posuere erat a ante venenatis dapibus posuere velit aliquet</li>
+  <li>Vivamus sagittis lacus vel augue laoreet rutrum faucibus dolor auctor.</li>
+  <li>Vestibulum id ligula porta felis euismod semper.</li>
+  <li>Integer posuere erat a ante venenatis dapibus posuere velit aliquet.</li>
 </ol>
 
 <ul class="list">


gemmaleigh:
Hi @StephenGill, thanks for these.

I think it would be useful to have some information as to why numbered lists should be full sentences.

From the style guide:
Use numbered steps instead of bullet points to guide a user through a process.
Each step ends in a full stop because each step should be a complete sentence.

Numbered lists currently look like this:

typography gov uk elements

I can add an example of a series of steps, like the one below:

steps

Let me find out if numbered lists should only be used for a series of steps, or if there is a need for both styles of numbered lists.

StephenGill:
Hi Gemma,

Can't think of a reason to use <ol> apart from telling someone about a
series of steps.

I think can't think of a good reason why there are two different styles -
maybe it's just because the mainstream and whitehall publishing apps work
differently?

So ... I think it's a good idea to replace the current <ol> example with a
mainstream-style 'series of steps'.



On 31 May 2016 at 10:13, Gemma Leigh <notifications@github.com> wrote:

> Hi @StephenGill <https://github.com/StephenGill>, thanks for these.
>
> I think it would be useful to have some information as to why numbered
> lists should be full sentences.
>
> From the style guide:
> Use numbered steps instead of bullet points to guide a user through a
> process.
> Each step ends in a full stop because each step should be a complete
> sentence.
>
> Numbered lists currently look like this:
>
> [image: typography gov uk elements]
> <https://cloud.githubusercontent.com/assets/417754/15669196/ed92a14a-2716-11e6-89b7-3957f1ca221d.png>
>
> I can add an example of a series of steps, like the one below:
>
> [image: steps]
> <https://cloud.githubusercontent.com/assets/417754/15669290/875bbf64-2717-11e6-9da0-fafbc08e1d39.png>
>
> Let me find out if numbered lists should only be used for a series of
> steps, or if there is a need for both styles of numbered lists.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/alphagov/govuk_elements/pull/238#issuecomment-222634306>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AFHgv25954EFkxXUAfx_IZ7KQ5LtbQkAks5qG_vSgaJpZM4IqDdH>
> .
>



-- 

*Stephen Gill*
Content Designer
Government Digital Service

T: 077962 98949

cjforms:
@gemmaleigh @StephenGill 
Have you got any examples from services where they've used ordered lists?
Have we got any user research to support the 'full sentences' advice in the content style guide?

I'd definitely support the removal of the long lorem-ipsum examples we've got now. 
Less sure that there are definitely no other needs for ordered lists. 

(Example that pops into my head: 
Here is the order of successiion for the UK:
1. HRH Prince Charles
2. HRH Prince William
3. HRH Prince George
... etc. 
I admit it might not be a very good example. I'll keep a look out for others.

StephenGill:
@cjforms @gemmaleigh

That example looks fair enough to me - and I can't see any downside in supporting that style as well if people think it would work better in the context of their service.

I'm not aware of any research to support of the complete sentences thing: I suspect it's something we've adopted because when we're describing steps in processes, we want to be able to write in multiple sentences and keep them short and simple - rather than use a single, long sentence with multiple clauses.

cjforms:
I don't disagree with the style guide, the advice seems good to me. My concern is only that we don't want to exclude shorter list items until we know that services don't need them.

views/guide_typography.html
```diff
@@ -134,7 +134,7 @@ <h3 class="heading-medium" id="typography-links">Links</h3>
 
   <h3 class="heading-medium" id="typography-lists">Lists</h3>
   <p class="text">
-    List items start with a lowercase letter and have no full stop at the end.
+    Bulleted list items start with a lowercase letter and have no full stop at the end. Numbered list items are written as full sentences, so they start with an uppercase letter and have a full stop at the end.
```
36degrees: Given the conversation in the other PR, can I suggest rather than duplicating we just defer to the style guide here, like we do in the markdown documentation?

e.g.

> See the [style guide](https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#bullet-points-steps) for guidance on how to punctuate lists.

We'd still update the examples.

36degrees:
Superseded by #454

Issue #239, opened 2016-05-31
The <abbr> tag and style


trevorsaint:
What is the desired style of this tag?

In Chrome we have no underline and in Firefox we have a dotted border bottom. Because no styling has been applied other than cursor: help; I am under the impression this is how GDS want it.

Has this been missed? Should I set text-decoration: none on this element?

Any steer would be good.

Thanks

gemmaleigh:
Hi Trevor,

The only example I can find of `<abbr>` styling is in the [GOV.UK component guide](http://govuk-component-guide.herokuapp.com/components/govspeak/fixtures/statistic_headlines/preview). 

This uses `cursor:help` which shows ? on hover and reveals the value of the title attribute.
`<abbr title="Departmental Expenditure Limit">DEL</abbr>`

`text-decoration: none` is also being applied, this isn't being copied to the reset used by GOV.UK elements.

For now, I'd use `text-decoration: none` and I'll add it to a list of things we should provide an example of.

trevorsaint:
Hi Gemma. I don't need an example of it :). It was an observation. I would have added the styling myself, but wanted to know why it had not been added into Gov Elements. I will add this myself in the hope of removing it when it is in your reset file. Thanks.

robinwhittleton:
@trevorsaint so the history is that `abbr` used to be styled in all browsers with `border-bottom: 1px dotted`. In W3’s [default HTML styles](https://www.w3.org/TR/html51/rendering.html#non-replaced-elements-phrasing-content) they suggest `text-decoration: dotted underline` instead. Firefox switched their user agent styles to match this when they added support for dotted underlines, but Chrome don’t so stuck with a dotted border. Our reset stylesheet on elements is out of date (on of the reasons I dislike them, but that’s another matter) so only resets border. We also need to reset the `text-decoration-style` ([as per the rest of GOV.UK](https://github.com/alphagov/static/commit/3042caaa277106f1493aeb60466239948ad4c8a4)), so I’ll do a PR for that now.

PR #241, opened 2016-06-01
Remove abbr underlines in Firefox

robinwhittleton:
Firefox now supports text-decoration with a dotted underline; consequently (and inline with W3’s recommended default stylesheet) Mozilla updated Firefox’s default stylesheet to style abbrs and acronyms with a dotted text-decoration instead of a dotted border. Our reset stylesheet removes borders from abbr and acronym elements, but doesn’t reset text decoration. The upshot of all this is that every abbr on styled by govuk_elements now has dots underneath it.

This patch simply adds a rule to remove text-decoration from abbr and acronym elements. If text-decoration is being deliberately set anywhere it’ll override this base rule.

### DIFF
--- a/public/sass/elements/_reset.scss
+++ b/public/sass/elements/_reset.scss
@@ -117,3 +117,7 @@ td {
   line-height: inherit;
   font-weight: normal;
 }
+
+abbr[title], acronym[title] {
+  text-decoration: none;
+}


robinwhittleton:
Fixes https://github.com/alphagov/govuk_elements/issues/239

gemmaleigh:
👍 thanks @robinwhittleton!

public/sass/elements/_reset.scss
```diff
@@ -117,3 +117,7 @@ td {
   line-height: inherit;
   font-weight: normal;
 }
+
+abbr[title], acronym[title] {
```
nickcolley: Could this be achieved without the title attribute selectors? This would enable people to style an abbr or acronym by putting a selector directly on it.

gemmaleigh: This won't prevent anyone from styling it by adding a class, this will just override the browser's default behaviour - which is to add a dotted underline or border.

eg `<abbr title="Web Content Accessibility Guidelines">WCAG</abbr>`

The title attribute is optional. 
I'm not sure we'd want abbreviated words without a title to be underlined either.

Normalize does this:
https://github.com/necolas/normalize.css/blob/master/normalize.css#L115

```
/**
 * 1. Remove the bottom border in Firefox 39-.
 * 2. Add the correct text decoration in Chrome, Edge, IE, Opera, and Safari.
 */

abbr[title] {
  border-bottom: none; /* 1 */
  text-decoration: underline; /* 2 */
  text-decoration: underline dotted; /* 2 */
}

```

.

@robinwhittleton, should we also remove the border from abbr without an title and remove the bottom border as Normalize also does for older versions of Firefox?

nickcolley: I was thinking along the lines of http://jsbin.com/vejeqaxiwu/edit?html,css,output but I guess utilities would probably be forced anyway.

PR #242, opened 2016-06-01
Add grunt lint to the test task

gemmaleigh:
When `grunt test` is run, run the scss lint task first and then check the
app runs.

Update the output.


gemmaleigh:
Closing as this isn't working, it should fail with error.

PR #243, opened 2016-06-01
Fix scss lint errors

gemmaleigh:
Disable the DuplicateProperty property linter in `_shame.scss`, to allow
the second property fall back to the first if not supported.

cc. @robinwhittleton.


robinwhittleton:
I suppose technically the second is the fallback, but yes, I forgot to lint ❌

PR #259, opened 2016-06-14
Use Nunjucks for templating

gemmaleigh:
This pull request:
 - adds the Jinja version of GOV.UK template as a dependency
 - removes Mustache libraries, and related Grunt tasks
 - converts all templates to use Jinja syntax

[This aims to make GOV.UK elements work in the same way as the prototype kit](https://github.com/alphagov/govuk_prototype_kit/pull/123), 
to make it easier to share HTML between the two.

cc. @joelanman @timpaul @dsingleton.


joelanman:
looks good to me, this is a lot of work, and it's a big upgrade for Elements - great job!

joelanman:
I won't merge because I'm not confident in the "include Elements as a package" thing, so I'm not able to test that side of it

gemmaleigh:
That should be fine, the npm module is only concerned with what's in `/public/sass/`

timpaul:
NICE ONE GEMMA LEIGH

PR #260, opened 2016-06-15
Create a github folder with sample issue and pull request templates

gemmaleigh:
Here is an issue template and a pull request template, slightly
modified from the choose your own adventure Open Source templates:
https://www.talater.com/open-source-templates/#/.

The idea for this came from the article here: http://alistapart.com/article/commit-to-contribute.


joelanman:
This is cool and will probably be helpful. I think we should be careful not to introduce too much friction, especially on Issues. I think it'd be best to make the language as simple as possible, and ask for as little as possible, so people aren't intimidated - GitHub is already daunting for many people.

timpaul:
Yeah, I read that article earlier and figured we should do this (also, loving the choose-your-own-adventure site). Two thoughts:

1. If we create a .github/ folder should we also put the .CONTRIBUTING file in it?
2. In the PR template we refer to our 'code style'. Is this documented, and if so, should we link to that?

joelanman:
@timpaul we have some code style guides here: https://github.com/alphagov/styleguides

PR #261, opened 2016-06-18
Fix default text for global cookie message

gemmaleigh:
This pull request is another step towards getting the GOV.UK elements app to work in the same way as the prototype kit. 🎆 

Add a config file, ensure that server.js uses the config file, then define a cookieText variable and pass the cookieText variable to all views. 

Also add the missing blocks for the cookie message in both layout files.

Screenshot before (empty blue cookie banner at the top of the screen):

gov uk elements - before

Screenshot after (cookie banner with text at the top of the screen):

gov uk elements - after

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -1,0 +1,10 @@
+// Use this file to change prototype configuration.
+
+// Note: this config can be overridden using environment variables (eg on heroku)
+
+module.exports = {
+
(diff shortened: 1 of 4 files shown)


gemmaleigh:
cc. @joelanman

joelanman:
looks good!

PR #264, opened 2016-06-21
Add a link to the GDS front end survey

gemmaleigh:
Add link to the survey to the top of the README.

cc. @timpaul.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,28 @@
 GOV.UK elements
 ===============
 
+---
+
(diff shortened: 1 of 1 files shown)


joelanman:
looks good to me, though maybe the call to action could be emphasised? That link could be bold, and we could remove the link from GOV.UK, so the survey's now the only link at the top

timpaul:
Now it's in situ - I think we need something much more compact, like:

**You can help us improve this app by completing our [5 minute survey](#)**

Could we add something across the top of the Elements web pages too?

PR #268, opened 2016-06-23
Use the button style to prompt for a file upload

### LINKED ISSUE #202: File upload pattern
It would be good to have a pattern for this, I see Digital Marketplace has this

http://alphagov.github.io/digitalmarketplace-frontend-toolkit/forms/upload.html

gemmaleigh:

This fixes #202.

### DIFF
--- a/app/views/snippets/form_file_upload.html
+++ b/app/views/snippets/form_file_upload.html
@@ -1,0 +1,4 @@
+<label class="button">
+  Upload a file
+  <input type="file" class="hidden" />
+</label>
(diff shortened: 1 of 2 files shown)


app/views/snippets/form_file_upload.html
```diff
@@ -0,0 +1,4 @@
+<label class="button">
+  Upload a file
+  <input type="file" class="hidden" />
```
robinwhittleton: Out of interest, why `hidden` and not `visually-hidden`?

Issue #151, opened 2016-01-05
Hide code snippets when JavaScript is disabled.


gemmaleigh:
[Prism.js](http://prismjs.com/) is used to format and style the code snippets. 

When JavaScript is disabled, hide the code snippets.

timpaul:
Is there any way of escaping the snippets so they still output without JavaScript? No worries if not - given that people can always view source it's not essential to offer snippets to non-JS users.

joelanman:
is it bad to show the snippets without formatting? The content might still be useful to people

PR #273, opened 2016-07-13
Fix code snippets when js is disabled

gemmaleigh:
This PR adds a Grunt task to generate encoded HTML snippets, and uses those within the `<code>` example blocks. 

This could do with some finessing, but is a quick-fix for now. It was spotted by Richard Morton yesterday and relates to the closed issue #151 (better to have the code snippets shown at all times, than hidden when JS is disabled).
 
I'd add an issue as a reminder to improve the styling of these snippets in future. 

Screenshots below.

When JS is disabled, code snippets underneath each example render like so:
screen shot 2016-07-13 at 11 45 30

This PR ensures that when JS is disabled - the code snippet is still shown:

screen shot 2016-07-13 at 11 46 59

cc. @cfq


joelanman:
very cool :)

cfq:
LGTM 👍  :-)

PR #283, opened 2016-08-11
Include the shim-links-with-button-role.js script from the frontend toolkit

gemmaleigh:
Include a script from the govuk frontend toolkit to shim anchors with
`role="button"` to be 'clicked' via the space key.

For context, here is the PR adding this script to the govuk_frontend_toolkit.
https://github.com/alphagov/govuk_frontend_toolkit/pull/310


gemmaleigh:
cc. @robinwhittleton @paulmichaelsmith. 

We should add documentation to the toolkit too.
https://github.com/alphagov/govuk_frontend_toolkit/blob/master/docs/javascript.md

paulmsmith:
I can look at that tomorrow or Wednesday.

paulmsmith:
+1

PR #285, opened 2016-08-17
Add `!default` flag to image `$path`

gavboulton:
The `!default` flag will only update the value of the variable when it has not been set. Adding this flag makes it easier to set when including elements modules in project specific files.


gemmaleigh:
Thanks @gavboulton, I'll check this out.

robinwhittleton:
Nice catch! Let’s merge.

PR #286, opened 2016-08-18
Form validation patterns for conditionally revealing content

### LINKED ISSUE #249: Form validation patterns for Conditionally revealing content
I'm looking at form validation patterns.

It would be good for a collection of validation patterns which incorporate conditionally revealing content.

As an example:
ni-number-error
 
The first validation in this case would be: Choose an answer

Secondly if yes has been chosen we now have the error on the National Insurance number and not the answer.

ni-number-input-error

It would be interesting to see how you handle the styling aspects of this and if it has been considered.

Thanks

gemmaleigh:
This PR adds two more examples to the form elements example page.

This fixes the issue with inputs using the `.form-control` class inheriting a red border, when a parent `.error` class is used.

Before:

2 example form elements form elements gov uk elements

After:

There is an example of this working here:
http://govuk-elements-test.herokuapp.com/form-elements/example-form-elements/#example-conditional-errors

cc. @trevorsaint, @gavboulton, @robinwhittleton. 

This fixes #249.

### DIFF
--- a/app/views/examples/example_form_elements.html
+++ b/app/views/examples/example_form_elements.html
@@ -540,6 +540,11 @@
       </div>
 
       <div class="form-section">
+        <!-- Radio buttons (select one option to reveal content) with error -->
+        {% include "snippets/form_inset_radios_show_errors.html" %}
+      </div>
+
+      <div class="form-section">
         <!-- Checkboxes (select more than option to reveal content) -->
         {% include "snippets/form_inset_checkboxes.html" %}
       </div>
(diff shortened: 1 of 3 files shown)


gavboulton:
This looks good to me.

One small thing in the example, should "Error message about monthly pay goes here" go above the label text to match the current patterns.

gemmaleigh:
The error message should sit underneath the label text and before the form input, or even better, inside the label - to ensure it is read out by assistive technology. 

http://govuk-elements.herokuapp.com/errors/#highlight-errors

gavboulton:
I must have been looking at something else, you're right, it's correct.

gemmaleigh:
Hi @robinwhittleton, 👀  on this one please, I think it's good to go.

gavboulton:
:+1:

PR #290, opened 2016-08-22
Lint codebase using standard

gemmaleigh:
This PR updates GOV.UK elements to use [JavaScript Standard Style ](http://standardjs.com/). 

This work has [already been carried out for the GOV.UK prototype kit](https://github.com/alphagov/govuk_prototype_kit/pull/224):

[Standard](https://www.npmjs.com/package/standard) is an opinionated JavaScript linter that uses [eslint](http://eslint.org/) underneath and acts as a set of sensible defaults for it; [here are the most important rules](https://github.com/feross/standard#rules).

Ideally any shared JavaScript will be moved to the govuk_frontend_toolkit in the future.

TODO:

- [x] Update CONTRIBUTING.md
- [x] Lint details.polyfill.js (previously ignored by line 40 of package.json)

cc. @tvararu, @nickcolley.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,6 @@
     "ignore": [
       "lib",
       "public/javascripts/vendor/",
-      "node_modules",
       "tests"
     ]
   },
--- a/package.json
+++ b/package.json
@@ -31,13 +31,13 @@
     "readdir": "0.0.6",
     "standard": "^7.1.2"
   },
-  "scripts": {
+  "scripts":{
     "test": "standard && ./node_modules/grunt-cli/bin/grunt test"
   },
   "standard": {
     "ignore": [
       "lib",
-      "public",
+      "public/javascripts/vendor/",
       "node_modules",
       "tests"
     ]
--- a/package.json
+++ b/package.json
@@ -28,10 +28,19 @@
     "grunt-sass": "^1.0.0",
     "grunt-shell": "^1.3.0",
     "node-sass": "^3.2.0",
-    "readdir": "0.0.6"
+    "readdir": "0.0.6",
+    "standard": "^7.1.2"
   },
   "scripts": {
-    "test": "./node_modules/grunt-cli/bin/grunt test"
+    "test": "standard && ./node_modules/grunt-cli/bin/grunt test"
+  },
+  "standard": {
+    "ignore": [
+      "lib",
+      "public",
+      "node_modules",
+      "tests"
+    ]
   },
   "bugs": {
     "url": "https://github.com/alphagov/govuk_elements/issues"
--- a/public/javascripts/application.js
+++ b/public/javascripts/application.js
@@ -1,161 +1,142 @@
-function ShowHideContent() {
-  var self = this;
+/* global $ */
+/* global jQuery */
+/* global GOVUK */
 
+function ShowHideContent () {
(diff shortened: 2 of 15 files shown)


public/javascripts/application.js
```diff
@@ -1,161 +1,142 @@
-function ShowHideContent() {
-  var self = this;
+/* global $ */
+/* global jQuery */
+/* global GOVUK */
```
tvararu: Note that an alternative way of avoiding this is doing:

```js
var $ = window.$
var GOVUK = window.GOVUK
```

Inside the code.

This works fine though. 👌

nickcolley: I prefer this because it is closer to what we'd do with a 'proper' module system in the future, wherein we have no globals.

server.js
```diff
@@ -1,53 +1,53 @@
...
-    routes = require(__dirname + '/app/routes.js'),
-    app = express(),
-    port = (process.env.PORT || 3000);
+var path = require('path')
+var express = require('express')
+var nunjucks = require('express-nunjucks')
+var routes = require(path.join(__dirname, '/app/routes.js'))
+var app = express()
+var bodyParser = require('body-parser')
+var config = require(path.join(__dirname, '/app/config.js'))
```
tvararu: `path.join`s in `require` calls should not be necessary, they should already be cross-platform compatible.

[I didn't do them on govuk_prototype_kit.](https://github.com/alphagov/govuk_prototype_kit/pull/224/files#diff-78c12f5adc1848d13b1c6f07055d996eR5)

gemmaleigh: Thanks! @tvararu, I think that was just overzealous copy and pasting on my part.

public/javascripts/govuk/details.polyfill.js
```diff
@@ -6,188 +6,185 @@
```
tvararu: I think we need to update this script tag to reference the new location: https://github.com/alphagov/govuk_elements/blob/32553af219c7b9fbad74aa71feebb32f38bcf643/app/views/includes/scripts.html#L9

It's 404ing when I run this new updated branch locally.

package.json
```diff
@@ -28,10 +28,19 @@
...
-  "scripts": {
-    "test": "./node_modules/grunt-cli/bin/grunt test"
+  "scripts":{
+    "test": "standard && ./node_modules/grunt-cli/bin/grunt test"
+  },
+  "standard": {
+    "ignore": [
+      "lib",
+      "public/javascripts/vendor/",
+      "node_modules",
```
tvararu: I *think* you don't need "node_modules" here but having it doesn't do any harm either.

tvararu:
This PR makes me very happy. 👏 

Gave it a look and ran it locally, apart from that polyfill thing I think this looks good. 👍

server.js
tvararu: It was our use of `__dirname + str` which causes the linter to recommend `path.join` instead, but it turns out in this use case neither the `__dirname` nor the `path.join` is necessary.

public/javascripts/application.js
robinwhittleton: This is being moved over to frontend_toolkit anyway so this is a temporary fix.

PR #292, opened 2016-08-24
Remove the $path override from the helpers partial

gemmaleigh:

This PR fixes image paths for GOV.UK elements, which are currently broken.

## What does it do?

Pull #285 added a `!default` flag to the $path variable, as `$path` is already set as `false !default` in url helpers, the file-url function fell through to use image-url. 
This resulted in broken images for GOV.UK elements.

## How has this been tested?

I think the simplest solution here is to not try and define `$path` in the govuk-elements-sass files, instead this should be defined in the main.scss file for each project.

Moving the `$path` variable to main.scss for govuk elements fixes this. It will however, break background images for anyone using the govuk-elements-sass files, they will need to define $path for their own project.

Although it is a breaking change, I think it's a sensible one.

cc. @gavboulton can you let me know if this solution works for you, as you introduced this change?

cc. @robinwhittleton.

This will need documentation to make it clear that $path must be defined on a project by project basis. [Digital marketplace already do this here](https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/master/app/assets/scss/application.scss#L7).

Broken background images before:

layout gov uk elements -1

Working background images after:

layout gov uk elements

## What type of change is it?

- [x] Breaking change (fix or feature that would cause existing functionality to change)

- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,9 +11,9 @@ GOV.UK elements
 
 GOV.UK elements is three things:
 
-1. an online design guide, explaining how to make your service look consistent with the rest of GOV.UK
+1. [an online design guide](http://govuk-elements.herokuapp.com/), explaining how to make your service look consistent with the rest of GOV.UK
 2. an example of how to use the code in the [GOV.UK template](https://github.com/alphagov/govuk_template) and the [GOV.UK frontend toolkit](https://github.com/alphagov/govuk_frontend_toolkit)
-3. an npm package of the Sass files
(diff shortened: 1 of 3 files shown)


gavboulton:
This looks good to me.

Sorry everyone for breaking your things!

gemmaleigh:
@gavboulton no worries! The override in helpers.scss was just confusing matters, it's much simpler if it works in this way.

PR #294, opened 2016-08-25
Add a new section on external links

gemmaleigh:

This pull request fixes the layout of the "Links" section (see screenshot).

## What does it do?

This pull request adds a new section for external links. This is to draw attention to external link styles, which have been deprecated. 

It uses the correct pattern to highlight information on a page and creates a new snippet for external links (so it is clear how these differ from links in body copy).

Before:

typography gov uk elements

After:

typography gov uk elements

## What type of change is it?

### DIFF
--- a/app/views/guide_typography.html
+++ b/app/views/guide_typography.html
@@ -133,14 +133,15 @@
 </code>
 </pre>
 
-  <h3 class="heading-medium" id="typography-external-links">External links</h3>
+  <h4 class="heading-medium" id="typography-external-links">External links</h4>
 
   <div class="panel panel-border-wide text">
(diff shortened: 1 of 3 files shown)


edwardhorsford:
Could I suggest removing the bold from the text? We advise services to avoid bold where possible, particularly in long blocks of text.

joelanman:
maybe leave the first line bold? otherwise it could easily be ignored

gemmaleigh:
I've updated the screenshots, let me know if that is better.

joelanman:
Lgtm

nickcolley:
👍

PR #295, opened 2016-08-26
[WIP, Do not merge] Update prism and increase the contrast for code snippets

gemmaleigh:

This PR is needed to increase the contrast for the code snippets.
At present, the background panel is very light. 

I'm not suggesting using the darker grey in the "After" screenshot, but it would be good to have a discussion about the colours we can use to fix this. 

Before:
layout gov uk elements

After: 
colour - after - layout gov uk elements

## What type of change is it?


nickcolley:
We could think about having a dark background for example



https://atom.io/themes/one-dark-ui

edwardhorsford:
@gemmaleigh It's not clear what issue you're trying to address. Is it that the box the code sits in isn't distinct enough?

selfthinker:
Shouldn't prism.scss be in a "vendor" folder as well? Now it looks like it might ours but isn't following our indentation style (tabs instead of 2 spaces). Having it in a "vendor" folder will also make linting rules easier.
(I know the placement of the file is not part of this PR but the changes to the indentation style is. That's why I noticed it. And it might generally be a good idea to move it.)

PR #296, opened 2016-08-27
Custom radios / checkboxes

robinwhittleton:
This requires no change to the existing HTML: the custom inputs are built using pseudoelements. In a no-JS / no-CSS / IE8 scenario it falls back to standard inputs. To style the radio and checkboxes differently a class is needed on the label - this is automatically added by the new SelectionButton JS.

This is partially based on Verify’s styles, and updates a design by @timpaul and @joelanman inspired by observations from user research and lots of requests from @cjforms. The inputs are sized to match the height of text inputs.

Image showing the design of the new radio buttons and checkboxes

Tested on:
- IE 7-11
- Edge 14
- OS X Firefox 48
- OS X Chrome 51
- OS X Safari 9
- iOS Safari 5.0 - 9.1  
- Android 4.4+ Chrome

Android Browser on BrowserStack’s Nexus 4 (4.2) is massively unresponsive and doesn’t seem to let you click on the radios most of the time. It will need to be tested on a real device.

Would absolutely love any feedback.
- [x] Base implementation + testing
- [x] Check that this works with click events in Selenium (testing in verify indicated possible problems)
- [x] Make it use `currentColor` for compatibility with custom colour themes
- [x] Make the pseudo-elements permanent and just repaint the background colour on state change
- [x] Add a translucent hover state
- [x] Test in old Android devices
- [x] Fix general problems with Dragon Naturally Speaking 13 on IE11 (Dragon 12.5 is fine)
- [x] Padding no longer exists after legends - investigate recommended changes

### DIFF
--- a/public/sass/elements/_forms.scss
+++ b/public/sass/elements/_forms.scss
@@ -145,9 +145,14 @@ textarea {
   @include core-19;
   width: 100%;
 
-  padding: 4px;
-  background-color: $white;
-  border: 2px solid $grey-1;
+  padding: 5px 4px 4px;
+  background-color: transparent;
+  border: 2px solid;
+  border-color: currentColor;
+
+  @include ie-lte(8) {
+    border-color: $text-colour;
+  }
 
   // TODO: Remove 50% width set for tablet and up
   // !! BREAKING CHANGE !!
--- a/public/sass/elements/_forms.scss
+++ b/public/sass/elements/_forms.scss
@@ -148,11 +148,6 @@ textarea {
   padding: 5px 4px 4px;
   background-color: transparent;
   border: 2px solid;
-  border-color: currentColor;
-
-  @include ie-lte(8) {
-    border-color: $text-colour;
-  }
 
   // TODO: Remove 50% width set for tablet and up
   // !! BREAKING CHANGE !!
--- a/public/sass/elements/_helpers.scss
+++ b/public/sass/elements/_helpers.scss
@@ -22,8 +22,7 @@
 }
 
 // Hide, but not for screenreaders
-.visually-hidden,
-.visuallyhidden {
+@mixin visually-hidden {
   position: absolute;
   overflow: hidden;
   clip: rect(0 0 0 0);
@@ -32,4 +31,9 @@
   margin: -1px;
   padding: 0;
   border: 0;
+}
+
+.visually-hidden,
+.visuallyhidden {
+  @include visually-hidden;
 }
--- a/public/sass/elements/forms/_form-block-labels.scss
+++ b/public/sass/elements/forms/_form-block-labels.scss
@@ -84,6 +84,8 @@
       position: absolute;
       top: 10px;
       left: 8px;
+      -moz-transform: rotate(-45deg); // Firefox 15 compatibility
+      -o-transform: rotate(-45deg); // Opera 12.0 compatibility
       -webkit-transform: rotate(-45deg); // Safari 8 compatibility
       -ms-transform: rotate(-45deg); // IE9 compatibility
       transform: rotate(-45deg);
--- a/public/sass/elements/forms/_form-block-labels.scss
+++ b/public/sass/elements/forms/_form-block-labels.scss
@@ -96,6 +96,12 @@
       &:hover::after {
         @include opacity(.15);
       }
+      // hide the hover state in iOS / mobile Chrome
+      @media (hover: none) {
+        &:hover::after {
+          opacity: 0;
+        }
+      }
     }
 
     // Focused state
--- a/public/sass/elements/forms/_form-block-labels.scss
+++ b/public/sass/elements/forms/_form-block-labels.scss
@@ -9,13 +9,15 @@
   clear: left;
   position: relative;
 
-  padding: (7px $gutter-one-third 7px $gutter * 1.5);
-  margin-bottom: $gutter-half;
+  padding: (8px $gutter-one-third 9px 50px);
+  margin-bottom: $gutter-one-third;
 
   cursor: pointer; // Encourage clicking on block labels
 
   @include media(tablet) {
     float: left;
+    padding-top: 7px;
+    padding-bottom: 7px;
     // width: 25%; - Test : check that text within labels will wrap
   }
 
@@ -25,8 +27,8 @@
     cursor: pointer;
     left: 0;
     top: 0;
-    width: 37px;
-    height: 37px;
+    width: 38px;
+    height: 38px;
 
     // IE8 doesn’t support pseudoelements, so we don’t want to hide native elements.
     @if ($is-ie == false) or ($ie-version == 9) {
@@ -42,8 +44,8 @@
       border: 2px solid currentColor;
       border-radius: 50%;
       background: transparent;
-      width: 33px;
-      height: 33px;
+      width: 34px;
+      height: 34px;
       position: absolute;
       top: 0;
       left: 0;
@@ -53,11 +55,11 @@
       content: "";
       border-radius: 50%;
       background: transparent;
-      width: 23px;
-      height: 23px;
+      width: 20px;
+      height: 20px;
       position: absolute;
-      top: 7px;
-      left: 7px;
+      top: 9px;
+      left: 9px;
     }
 
     &.selection-button-radio.selected::after {
@@ -68,8 +70,8 @@
       content: "";
       border: 2px solid currentColor;
       background: transparent;
-      width: 33px;
-      height: 33px;
+      width: 34px;
+      height: 34px;
       position: absolute;
       top: 0;
       left: 0;
@@ -78,13 +80,13 @@
     &.selection-button-checkbox::after {
       content: "";
       border: solid transparent;
-      border-width: 0 0 4px 4px;
+      border-width: 0 0 5px 5px;
       background: transparent;
-      width: 15px;
-      height: 9px;
+      width: 17px;
+      height: 7px;
       position: absolute;
-      top: 9px;
-      left: 9px;
+      top: 10px;
+      left: 8px;
       -webkit-transform: rotate(-45deg); // Safari 8 compatibility
       -ms-transform: rotate(-45deg); // IE9 compatibility
       transform: rotate(-45deg);
@@ -107,7 +109,7 @@
 
   @include media (tablet) {
     margin-bottom: 0;
-    margin-right: $gutter-half;
+    margin-right: $gutter;
   }
 }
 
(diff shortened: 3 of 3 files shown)


joelanman:
Tim Paul and me :)

robinwhittleton:
Oops! Fixed.

paulmsmith:
Nice

nickcolley:
Might be worth marking as [Do not merge] since it relies on the upstream addition to selection buttons?

nickcolley:
Some interesting information around touch sizes http://www.lukew.com/ff/entry.asp?1085 , might be relevant :P

robinwhittleton:
Interesting research! The actual input has increased in height from 29px to 37px, but as we’ve removed the grey background it no longer needs the padding which means that the overall height of the hit area has gone down from 60px to 37px. As long as people are still trying to press on the control maybe that’s not a problem though.

joelanman:
this work comes from user research that shows that in general people don't know the grey area is the hit box. so while the actual hit box is reduced, the perceived hit box is increased.

also worth noting that, at least in Chrome on OSX, our current radios aren't actually 29px (perceived)

image

robinwhittleton:
Just remembered, I also want to get @cfq / @accessiblewebuk / @aduggin’s seal of approval on this before we merge, or at least before we do a release.

aduggin:
@robinwhittleton Can you make a rendered version available on a public url (for example, in the sandbox) so we can get it fully tested?

joelanman:
Why does it need to fall back to default inputs in a no js situation?

robinwhittleton:
`<input />` is an empty element; it can’t have pseudo-elements because there is text content for a `::before` or `::after` to be before or after. Therefore if we want to keep the same HTML structure we need to attach out pseudo-elements to the parent `label`. That poses a problem – we can’t style the parent based on child state ([currently](https://drafts.csswg.org/selectors-4/#relational)) so we need to base our pseudo-element styles on the classes that SelectionButtons applies. If we don’t have these classes then the fake inputs will be able to display their selected state, so we need to make sure that they don’t show up in the absence of JavaScript.

cjforms:
Based on a design by @timpaul and @joelanman inspired by observations from user research and lots of requests from me.

edwardhorsford:
Is this deployed anywhere we can try?

What do these do on hover?
What do these look like with focus?

Any thoughts on the below hover effect? (stolen from @selfthinker)
verify_radios_with_hover

robinwhittleton:
>Is this deployed anywhere we can try?

Not yet.

>What do these do on hover?

No change. I like @selfthinker’s idea. What would you want to do with checkbox hover states?

>What do these look like with focus?

Due to removal of padding I went for outline of the element (square with checkboxes):
Demo of outline

edwardhorsford:
@robinwhittleton It's something I'd want to try in user research. I suspect you could do a greyed out tick on hover. I think once the radio or checkbox has been checked, hovering over it should *not* show the greyed symbol.

An alternate would be to show a hover effect that is not the same as the tick or dot and could work whether the control was selected or not. Something like changing the white ring for a grey ring, or grey inner-stroke.

---

Seperately - how do these work for people who use css to adjust contrast / change colours?

robinwhittleton:
It’s worth having the conversation about contrast issues, but my gut feeling is that browser-default checkboxes and radios already don’t respond to setting user colours?

edwardhorsford:
@robinwhittleton My worry is especially about something like this happening:
screen shot 2016-08-30 at 12 13 09

Where the controls themselves disappear rather than being the wrong style.

Chatting with @nickcolley on Slack, it sounds like we may have a solution?

nickcolley:
The idea I put forward on slack was to remove the white background from the input and allow the colour to be defined by the text color of the label.

image
image

In the above examples the container I set the `background-color` and `color` and you can see the changes reflect properly when using `currentColor` and removing the background from the input.

Might be some edge cases I've not thought about though..

robinwhittleton:
Hmm. That generally works. I think on Verify start page we’d want to keep the white background though.

verify-radio-background

nickcolley:
Yeah I think the current implementation should work fine even if the user changes their background/color sorry!

edwardhorsford:
@robinwhittleton The current live verify radios do not work if the user has custom colours set (using firefox). Clicking on the option does nothing (visually). Where can I raise this?
screen shot 2016-08-30 at 14 07 54

robinwhittleton:
[New issue on verify-frontend](https://github.com/alphagov/verify-frontend/issues/new)? The fix might be to complete this PR though and merge that in.

tombye:
This is ace, thanks Robin. Minor question: is the `js-enabled` class needed in the selectors?

I might be missing something but I think the styles can all be based off the `selection-button-radio` and `selection-button-checkbox` classes which are only applied if the JS for that runs anyway. If that's the case it would be nice to hang these styles off that rather than being dependant on the user having JS at all.

robinwhittleton:
For the most part that wouldn’t be a problem, but imagine this situation:

1. Developer copies HTML from an existing service by opening devtools and right-click / Copy HTML on a node.
2. Developer then pastes that rendered HTML (which includes `selection-button-radio`) into a new place and changes the name and text of the radio and label.
3. Developer doesn’t notice any difference as the JS is applying that class anyway.
4. User without JS finds a custom radio button and no way of selecting it.

Obviously good developers should test with JS disabled, but it’s easy for something to slip through. Wrapping all of this in a `.js-enabled` is being a little more defensive.

tombye:
Makes sense.

edwardhorsford:
@robinwhittleton really interesting and useful - thanks for posting. I learn something new every day working here.

cjforms:
@robinwhittleton I did not realise that the removal of padding meant that the focus outline would shrink to the target. That's not so good. Although many people aim for the target - hence this whole large target idea - the majority do not (especially on mobile), so the focus ought to include the clickable area, especially including the label. Having the focus solely around the target gives the impression that only the target is clickable.

edwardhorsford:
@cjforms it would be good to do some research on this. I'm not sure it's as much of an issue - unless you're using keyboard controls, you don't normally see focus until *after* you've picked something. So the focus area isn't that important for indicating what you can click on. Right now the grey fill and border indicate the click area, and for me that's a bigger loss for indicating where you can pick.

"Although many people aim for the target - hence this whole large target idea - the majority do not (especially on mobile)" - do you mean the reverse? On mobile I agree that users pick the text (or the area in the grey box), but on desktop the vast majority I've observed go for the target.

edwardhorsford:
Can we try alternates with having the focus around the entire area? I'm not sure it does much for indicating click area, but if you're using keyboard controls it's probably an improvement.

cjforms:
@edwardhorsford Maybe I was wrong to quote 'the majority' and 'many'. I had these approximate ratios in mind: 
- more mobile users than desktop
- mobile users more likely to aim for clickable area than target
- I've seen roughly equal number of users aim for clickable area and for target.
But you've watched a lot more desktop users recently than I have, and I'm sure the proportions of users on mobiles and users on desktop vary a lot by service.
What I really meant was: making life easy for people who aim for the target is a good thing, and we also need to continue to make life easy for people who aim for the clickable area. 
+1 to your point about focus and keyboard controls. Let's test it.

robinwhittleton:
Had a meeting with @markhurrell / @joelanman / @gemmaleigh / @miaallers: general consensus what that we were happy with the approach but wanted to tweak the design a bit and add a hover state. I’ve done this and updated the image in the first post to match the current state.

selfthinker:
I prefer the smaller dot, so, I like the improvement.
Regarding the hover and having "stolen" from me, I just wanted to give a shout to @encosion who had the idea in the first place (at my previous job). Maybe he stole it from somewhere as well. ;-)

encosion:
Cheers @selfthinker! I don't recall but, there is nothing new under the sun...

nickcolley:
We should also check that the hover states do not affect devices that do simulated hovering.

https://github.com/twbs/mq4-hover-shim

robinwhittleton:
@nickcolley turns out we don’t need to shim CSS4 `hover: none` media queries as it’s supported in recent versions of Safari and Chrome.

nickcolley:
@robinwhittleton that'll work on newer iOS devices but we'd still find that bug in older devices if we tested them I think...

roc:
Looks great. I dispose of waste from mines and quarries mainly.

robinwhittleton:
Chatting with @adamsilver got me thinking about another problem. We know that a fair few people have `.js-enabled` but don’t get the rest of the JS (broken pipe to CDN / people MitMing the connection and messing with the JS / browser plugins breaking the environment etc). In this scenario the inputs will now be broken for them as they’ll never be able to select them, or at least see the visual state of them being checked. Is there anything we can do for these users?

adamsilver:
@robinwhittleton assuming the definition and initialisation happens in the same script (bundled/concat for prod etc) then you could do something similar to this:

	    <script>
	    	// Assuming the script is called GOVUK.SelectionButtons
	    	if(typeof GOVUK === 'undefined' || typeof GOVUK.SelectionButtons === 'undefined') {
	    	   // undo the incomplete enhancing
	    	}
	    </script>
	</body>

robinwhittleton:
Fixed that situation for some of the people in https://github.com/alphagov/govuk_template/pull/248, which will be released soon in govuk_template 0.18.2.

robinwhittleton:
Current status: waiting on @cfq / @accessiblewebuk to check these in Dragon NaturallySpeaking 13. DAC reported these working in 12.5 and failing in 13, so I want to get to the bottom of what’s going on before launching.

robinwhittleton:
Got a chance to sit down with Dragon 13 today and figured out what the problem is - it doesn’t recognise the original checkboxes / radios with our original hiding code. There’s a fix in https://github.com/alphagov/govuk_elements/commit/ea23ea5af2362d54d9eddfdadb19459bd5a71a28, and I’ve tested that in lots of different places. I haven’t yet tested in Dragon 12.5 though. Has anyone got a copy to test in? http://govuk-elements-wip.herokuapp.com/form-elements/#form-radio-buttons is currently running this latest code.

robinwhittleton:
@aduggin is going to run this back past the DAC to get the latest version checked out.

robinwhittleton:
With the [fix in place for Windows high contrast themes](https://github.com/alphagov/govuk_elements/pull/296/commits/6f6a1b0bca7d7583fd8c50b856d65cbc9a3d659c) DAC gave their thumbs up. I think, amazingly, that’s we’re ready to go.

robinwhittleton:
N.b. anyone using this will need to be on at least govuk_frontend_toolkit 4.17.0 as that contains the necessary JavaScript to get this to work properly.

gemmaleigh:
Thanks for all your work on this @robinwhittleton and to everyone else who contributed and offered advice.

We'll release an update to govuk_elements for custom radio buttons and checkboxes today and update the govuk_prototype_kit to include them too.

PR #304, opened 2016-09-07
Add bullet points describing use of disabled buttons

gemmaleigh:
Before:
screen shot 2016-09-07 at 15 14 32

After:
screen shot 2016-09-07 at 15 25 53

This fixes https://github.com/alphagov/govuk_elements/issues/201


robinwhittleton:
This presumably also invalidates #215?

PR #305, opened 2016-09-08
Add new section for file upload, with examples

### LINKED ISSUE #298: Upload a file button gets no focus
## How would you expect it to work?

I would expect the button to receive tabbed focus and a yellow border appear like all other form elements. The enter key or space bar may open up the file selection dialog.

## How does it work currently?

Skips tabbed focus completely.

## Feel free to suggest a fix...

Use javascript to set a focussed class on the button wrapper when the hidden input has focus.
Because the input is using display: none we won't be able to get focus so perhaps using opacity: 0 may work, or text-indent: -999px. I have not tried any of these approaches yet.

## Explain the steps to reproduce the bug

Tabbing through the DOM.

## Tell us about your environment

* Version used:
Latest version
* Browser Name and version:
Firefox 48.0.1
* Operating System and version (desktop or mobile):
Mac OSX El Captain
* Link to your project:
http://tfl-prosecutor.herokuapp.com/send-data-files

gemmaleigh:

This PR fixes #298, by reverting the custom file upload style. 

## What does it do?

The button style currently used prevents the button from gaining focus. 
This a bug, for now it is better to revert to the native file upload and in future find a custom solution which is accessible.

Before:
screen shot 2016-09-08 at 11 40 55

After:

screen shot 2016-09-08 at 11 39 43

After (focussed by tabbing through form examples page):

screen shot 2016-09-08 at 11 40 03

New section added to the "Form elements" guide:

screen shot 2016-09-08 at 11 40 19

## What type of change is it?

### DIFF
--- a/app/views/guide_form_elements.html
+++ b/app/views/guide_form_elements.html
@@ -145,6 +145,23 @@
 </code>
 </pre>
 
+  <div class="text">
+    <p>
+      We recommend using a native input using <code class="code">type="file"</code>, rather than a custom implementation.
+    </p>
+    <p class="lead-in">
+      This is so:
+    </p>
+    <ul class="list list-bullet">
+      <li>the control gains focus when tabbing through the page</li>
+      <li>the control functions using a keyboard</li>
+      <li>the control functions using assistive technology</li>
+      <li>the control functions when javascript is disabled</li>
+    </ul>
+    <p>
+      A custom implementation of this control would need to meet the criteria above.
+    </p>
+  </div>
 
   <h3 class="heading-medium" id="form-fieldsets">Fieldsets and legends</h3>
   <p class="text">
--- a/app/views/guide_form_elements.html
+++ b/app/views/guide_form_elements.html
@@ -28,6 +28,7 @@
         <li><a href="#form-focus-states">Form focus states</a></li>
         <li><a href="#form-hint-text">Hint text</a></li>
         <li><a href="#form-spacing">Spacing</a></li>
+        <li><a href="#form-file-upload">File upload</a></li>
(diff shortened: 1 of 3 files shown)


nickcolley:
Looks good, I wonder if we should put an explanation on why we are not doing a custom upload as I imagine some people will see this as an opportunity to re-implement it without realising the effort we need to go into to making sure a custom input behaves as well as a native input?

gemmaleigh:
I've added a section underneath to explain why we recommend using the native `input type="file"`.

Screenshot below:

explanation -form elements gov uk elements

nickcolley:
👍

app/views/guide_form_elements.html
```diff
@@ -145,6 +145,23 @@ <h3 class="heading-medium" id="form-file-upload">File upload</h3>
...
+      We recommend using a native input using <code class="code">type="file"</code>, rather than a custom implementation.
+    </p>
+    <p class="lead-in">
+      This is so:
+    </p>
+    <ul class="list list-bullet">
+      <li>the control gains focus when tabbing through the page</li>
+      <li>the control functions using a keyboard</li>
+      <li>the control functions using assistive technology</li>
+      <li>the control functions when javascript is disabled</li>
```
joelanman: should we generally refer to 'javascript not available', given [the majority case is not 'disabled'](https://gds.blog.gov.uk/2013/10/21/how-many-people-are-missing-out-on-javascript-enhancement/)

gemmaleigh: Yup, absolutely - feel free to open a PR to fix!

PR #306, opened 2016-09-12
Add npm start

gemmaleigh:
Ensure that 'npm start' can be used to start the govuk_elements application.

If no "start" property is specified on the "scripts" object, it will
run node server.js.
## What type of change is it?


nickcolley:
Could we change the scripts in package.json to look like

```json
  "scripts":{
    "test": "standard && grunt test",
    "start": "grunt"
  },
```
(package.json can get the binary globals from the local node_modules so you don't need to hard code them)

Then remove start.js?

nickcolley:
Then we could update https://github.com/alphagov/govuk_elements/blob/master/README.md#running-this-site-locally to use `npm start`

PR #309, opened 2016-09-12
Revert "Add npm start"

gemmaleigh:
Reverts alphagov/govuk_elements#306

This has broken the govuk_elements app, reverting to fix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,7 +177,7 @@ Install the required node modules
 
 Run the app
 
-    npm start
+    node start.js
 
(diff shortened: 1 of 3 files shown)


nickcolley:
I wonder if we could roll forward with changing 

https://github.com/alphagov/govuk_elements/blob/3aa9092185b5a018bd122d8270cdffc7a1f3e6e9/Procfile ?

gemmaleigh:
Closing in favour of #310,

PR #315, opened 2016-09-14
Ensure that a parent .form-group wraps each question

gemmaleigh:

This PR makes the examples more consistent, by ensuring that each question has a parent `.form-group` wrapper, so that validation styles can be applied. 

## What does it do?

In some examples the `.form-group` wrapper sat inside a fieldset. It is now used outside the fieldset, to match the "Errors and validation" examples.

## What type of change is it?


gemmaleigh:
cc. @robinwhittleton.

PR #319, opened 2016-09-19
Easier copying of HTML from code snippets

gemmaleigh:
## What does it do?

This PR adds `user-select`:`all` to preformatted code blocks, to make it easier to copy an entire code snippet. Clicking inside the code snippet selects the entire code block.

Before:
before-user-select-all

After:
user-select-all

## What type of change is it?


gemmaleigh:
Having had a discussion about this with @timpaul and @robinwhittleton, it's likely that users will want to copy only part of a code snippet - so this isn't helpful as-is.

PR #323, opened 2016-10-04
Fixed radio button jumping problem on Safari

PanosParalakis:
Updated block-label input styles

To fix radio button jumping bug on Safari browser

## What does it do?

It fixes the position
## How has this been tested?

Manually on Safari, Chrome
Executed grunt test-default task

Hard to catch the screenshot as it is on press down
## What type of change is it?

- [x ] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/public/sass/elements/forms/_form-block-labels.scss
+++ b/public/sass/elements/forms/_form-block-labels.scss
@@ -24,12 +24,12 @@
   // Absolutely position inputs within label, to allow text to wrap
   input {
     position: absolute;
-    top: 15px;
+    top: 21px; // 15px; updated as height is removed
     left: $gutter-half;
     cursor: pointer;
     margin: 0;
     width: 29px;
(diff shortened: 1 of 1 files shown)


robinwhittleton:
Thanks for the PR! Unfortunately while this potentially fixes the problem in Safari and Chrome, [they’re the two browsers on macOS that don’t resize the inputs visually based on their applied size](https://gdstechnology.blog.gov.uk/2015/08/27/making-radio-buttons-and-checkboxes-easier-to-use/). Either way, we’re looking to replace this style with [a new one for radios and checkboxes](https://github.com/alphagov/govuk_elements/pull/296), which is currently going through testing. That will replace all of this code, so it’s probably not worth your time coming up with a new version of this patch that works for everyone at this point. If you’d like to test out the new code I’d love your opinions.

PR #329, opened 2016-10-20
Bump govuk_frontend_toolkit to 4.18.4

gemmaleigh:
# 4.18.4

https://raw.githubusercontent.com/alphagov/govuk_frontend_toolkit/master
/CHANGELOG.md

- Lint codebase using standard
- Add semicolons at the start of IIFE's


robinwhittleton:
This has 4.18.2 as a dev dependency, and 4.18.4 as a dependency. We want to normalise that right?

nickcolley:
Any code that ends up in the final payload run by users should be in 'dependencies' any dependencies that are used to facilitate development like build tools etc should be in 'devDependencies'

PR #335, opened 2016-10-25
Remove guidance on external link icons

robinwhittleton:
As we’re removing these from frontend toolkit we can remove the guidance around their use.


gemmaleigh:
That's great, thanks @robinwhittleton!

PR #342, opened 2016-10-27
Remove rounded corners on form inputs elements and textareas in iOS

gemmaleigh:

This PR removes the rounded corners from form elements in iOS and also the inner shadow from form input elements and textareas.
## What does it do?

This PR doesn't affect select boxes (drop-down lists) as we'd like to keep native styles for this element.

Before:

realios_iphone_6s_plus

After:

realios_iphone_6s_plus 1
## What type of change is it?

Remove rounded corners for inputs and textarea elements using the
.form-control class.

Also set -webkit-appearance: none; to remove the inner shadow on these
elements.

### DIFF
--- a/public/sass/elements/_forms.scss
+++ b/public/sass/elements/_forms.scss
@@ -157,6 +157,8 @@ textarea {
 
 }
 
+// Allow a qualifying element, remove rounded corners from inputs and textareas
+// scss-lint:disable QualifyingElement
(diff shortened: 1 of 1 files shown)


robinwhittleton:
Checked this on various modern and old iOS / Android / IE Mobile, and on modern Safari on macOS and it all seems fine. Let’s merge.

Issue #343, opened 2016-11-09
Skiplink moves focus to before the phase banner


tombye:
The `<main>` tag currently wraps both the main content area and the phase banner. This means when moving through pages by keyboard only using the skiplink you have to pass the phase banner every time.

![govuk_elements_current_skiplink](https://cloud.githubusercontent.com/assets/87140/20145973/f5d11cd0-a699-11e6-816b-ba5d61168e21.png)

I suggest we replace the `<main>` tag with a wrapper div to establish the central column and start `<main>` after the phase banner instead, as a child of the wrapper div. This would allow users to skip directly to the page content.

![govuk_elements_proposed_skiplink](https://cloud.githubusercontent.com/assets/87140/20145985/fe6a2512-a699-11e6-8cae-fc46b4707065.png)

There would have to be some debate around whether or not to move the breadcrumbs outside of `<main>` too. www.gov.uk does this but I'm unsure it's the best approach as they can be specific to the page and the purpose of skiplinks is to allow keyboard users to avoid content repeated across pages.

robinwhittleton:
Yeah, I don’t agree that breadcrumbs should be outside `<main>`. For the phase banner specifically on elements, that should just be removed. At version 2.1.0 we’re definitely outside alpha. For the more general problem of a wrapper element, that’s presumably a problem that govuk_template will need to solve?

edwardhorsford:
To me it makes sense for it to be outside of main, or at the very least for the skip link to go _directly to the content_ and not to any breadcrumbs.

I wonder what @aduggin thinks.

aduggin:
I agree with Ed and Tom - skip links should take you directly to the main content.

It recommends this here: https://www.w3.org/TR/WCAG20-TECHS/G1.html
"The objective of this technique [skip links] is to provide a mechanism to bypass blocks of material that are repeated on multiple Web pages by skipping directly to the main content of the Web page"

and

"Example 2: A "Skip to main content" link
A Web page includes a variety of navigation techniques on each page: a bread crumb trail, a search tool, a site map, and a list of related resources. The first link on the page is titled "Skip to Main Content". A user activates the link to skip over the navigation tools."

gemmaleigh:
The removal of the phase banner from GOV.UK elements has fixed part of this issue, however, the breadcrumb links sit within the `<main>` content area.

[...]

PR #344, opened 2016-11-09
Remove the phase banner

robinwhittleton:
Elements definitely isn’t in alpha any more (2.1.0 currently), and the page that the survey link points at was last updated 3 years ago. Let’s just drop it.

Fixes the symptom (but not the underlying problem) of https://github.com/alphagov/govuk_elements/issues/343

### DIFF
--- a/app/views/guide_alpha_beta.html
+++ b/app/views/guide_alpha_beta.html
@@ -6,7 +6,6 @@
 
 <main id="content" role="main" tabindex="-1">
 
-  {% include "includes/phase_banner_survey.html" %}
   {% include "includes/breadcrumb.html" %}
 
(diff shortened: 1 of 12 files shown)


gemmaleigh:
It'a probably worth deleting the phase banner survey include as part of this PR too. It was added (as a duplicate of the original phase banner include) to get feedback for the govuk frontend alpha.

gemmaleigh:
https://github.com/alphagov/govuk_elements/blob/master/app/views/includes/phase_banner_survey.html

gemmaleigh:
This seems sensible, I've removed the survey phase banner include too. 

Thanks @robinwhittleton 💯

PR #348, opened 2016-11-14
Fix the spacing underneath the "inline" block label example

### LINKED ISSUE #345: Inconsistent spacing on conditional element in example
One of the examples includes a yes/no inline question, with conditional fields revealed once an answer is given.

The spacing on one of the conditional fields is wrong.

Video:
elements-yes-no-bug

## How would you expect it to work?
The correct spacing is on the 'no' answer - the yes answer has too little spacing.

## How does it work currently?
The spacing is inconsistent.

## Feel free to suggest a fix...
It looks like the second option is correctly spaced because it has a margin-top of 10px. However this margin top is applied only to the last child - in this case there are two children so the first hidden question does not get the margin.

## Explain the steps to reproduce the bug
Visit [govuk-elements example page](http://govuk-elements.herokuapp.com/form-elements/example-form-elements/) and try the yes / no inline example.

gemmaleigh:

This PR fixes #345.

## What does it do?

This PR ensures that when two block labels are floated next to each other (using the `.inline` class on a parent fieldset) that any conditionally revealed content - for example a panel revealing an additional question, has a top margin of 10px.

For example, a yes/no question:

```
<fieldset class="inline">
   <label class="block-label" data-target="the-revealed-panel">Yes</label>
   <div class="panel-border-narrow" id="the-revealed-panel">
      Additional content for "Yes" option
   </div>
</fieldset>
```

Ensure that this panel has a top margin of 10px, so the border doesn’t
connect with the block label above it.

fix-yes-no

## What type of change is it?

### DIFF
--- a/public/sass/elements/_panels.scss
+++ b/public/sass/elements/_panels.scss
@@ -54,7 +54,7 @@
 }
 
 // For inline panels
-.inline .form-group .panel-border-narrow {
+.inline .panel-border-narrow {
   margin-top: 10px;
   margin-bottom: 0;
 }


edwardhorsford:
Thanks Gemma!

robinwhittleton:
Well, it fixes this particular situation. I refer people to [line 43](https://github.com/alphagov/govuk_elements/blob/901c2d2aaf29691b5ede996ad00c7273eda7ca2d/public/sass/elements/_panels.scss#L43) though:

>`Note: This is incredibly fragile, and needs rebuilding.`

PR #359, opened 2016-11-21
Set color:inherit to match text and border colour for form controls

### LINKED ISSUE #358: Fix border colour on form controls
Borders of `.form-control`s are following the text colour, but that’s incorrectly defaulting to `#000`. Need to set `color: inherit;` for them so they correctly match the text colour.

gemmaleigh:

This PR fixes #358.

Borders on form controls have been inadvertently changed  to`#00000`, this is too dark.
These should remain the same colour as the form label text.

## What does it do?

As the color property isn't defined for `.form-control`, the user agent is setting this to `color: initial`, which is black #000000.

Setting `color:inherit` ensures the text inside a form control and the border colour become [$text-colour](http://govuk-elements.herokuapp.com/colour/#colour-sass-variables) `#0B0C0C`. 

Before:

before - form elements gov uk elements

After:

after - form elements gov uk elements

colour gov uk elements

## What type of change is it?

### DIFF
--- a/public/sass/elements/_forms.scss
+++ b/public/sass/elements/_forms.scss
@@ -146,6 +146,7 @@ textarea {
   width: 100%;
 
   padding: 5px 4px 4px;
+  color: inherit;
   background-color: transparent;
   border: 2px solid;
 


edwardhorsford:
@gemmaleigh does this also set the text colour within the inputs?

gemmaleigh:
Hi @edwardhorsford 

Yes, it sets the `color: inherit`, so the text colour within the inputs will inherit`$text-colour` (#0b0c0c), set on the body by [govuk_template](https://github.com/alphagov/govuk_template/blob/02ca6cafb81278582f65947a7d989f2323def554/source/assets/stylesheets/_basic.scss#L65).

Here's a bit more background (and how it affects the border colour too):

[CSS1 and CSS2 defined the initial value of the ‘border-color’ property to be the value of the ‘color’ property](https://www.w3.org/TR/REC-CSS1/#border-color). 

If no color value is specified for the border, the value of the 'color' property of the element itself will take its place.

[The initial value of border-color is currentColor](https://drafts.csswg.org/css-backgrounds-3/#border-color), this will match the color property if it exists, so that's why we're setting colour here.

As an aside, if the ‘currentColor’ keyword is set on the ‘color’ property itself, it is treated as ‘color: inherit’.

PR #363, opened 2016-11-25
New radios / checkboxes performance improvements

robinwhittleton:
Two small commits to improve the performance of the new radios and checkboxes.

1. remove the 300ms delay on mobile in some situations – @nickcolley can you review?
2. add the classes to recommended markup – @gemmaleigh can you review?

### DIFF
--- a/app/views/snippets/form_error_radio_show_errors.html
+++ b/app/views/snippets/form_error_radio_show_errors.html
@@ -37,14 +37,14 @@
 
       </legend>
 
-      <label class="block-label" for="personal_details_yes">
+      <label class="block-label selection-button-radio" for="personal_details_yes">
         <input id="personal_details_yes" type="radio" name="personalDetails" value="Yes">
-        <span></span> Yes, my personal details are correct
+        Yes, my personal details are correct
       </label>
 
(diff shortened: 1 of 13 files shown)


nickcolley:
Does this mean the state is still controlled by JavaScript?

If so this means that the first paint will be quick but there will be the same period of time where users will try to interact with the form and it will not respond.

robinwhittleton:
Yes, that’s true, but state post-interaction will be reflected in visual state when the JS does load. Which do you think is better?

app/views/snippets/form_error_radio_show_errors.html
```diff
@@ -37,12 +37,12 @@ <h1 class="heading-large">
 
       </legend>
 
-      <label class="block-label" for="personal_details_yes">
+      <label class="block-label selection-button-radio" for="personal_details_yes">
         <input id="personal_details_yes" type="radio" name="personalDetails" value="Yes">
         <span></span> Yes, my personal details are correct
```
gemmaleigh: Should these nested `<span>` elements also be removed? They have been removed in other examples.

PR #366, opened 2016-11-30
[For discussion] Examples of phase tags

gemmaleigh:

This example was missing from GOV.UK elements, add an example to show
how phase tags can be used outside of the phase banner.

alpha and beta banners gov uk elements

#### Put an `x` in all the boxes that apply

### DIFF
--- a/app/views/guide_alpha_beta.html
+++ b/app/views/guide_alpha_beta.html
@@ -23,6 +23,7 @@
       <ul class="list list-contents">
         <li><a href="#alpha-banner">Alpha banner</a></li>
         <li><a href="#beta-banner">Beta banner</a></li>
+        <li><a href="#phase-tags">Phase tags</a></li>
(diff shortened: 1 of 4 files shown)


edwardhorsford:
I think it would be good to have a sentence about when and how we would expect these to be used.

I think it may be clearer if there's only one screenshot, and only one html snippet. To me it's clear which bit of text you need to change.

gemmaleigh:
image

robinwhittleton:
Looks good, but it’ll need some text as to why you’d use this outside of a phase banner. Maybe something like:

> Phase tags are mostly used inside phase banners as an indication of the state of a project. It’s possible to use them outside phase banners, for example as part of a service header.

With appropriate links to Tim’s headers work.

robinwhittleton:
Will merge this next week unless I hear any dissent.

PR #373, opened 2016-12-06
Recognise Larry’s contribution to the Cabinet Office

robinwhittleton:
As [Chief Mouser to the Cabinet Office](https://www.gov.uk/government/history/10-downing-street#larry-chief-mouser) Larry plays a vital part in the day-to-day running of government. Given that we’re celebrating [100 years of the Cabinet Office this month](https://www.nationalarchives.gov.uk/cabinet-office-100) it makes sense to celebrate one of its most famous postings in a small way in this style guide.


app/views/guide_icons_images.html
```diff
@@ -70,7 +70,7 @@ <h3 class="heading-medium" id="images">Images</h3>
         <img src="/public/images/examples/3by2.jpg" alt="3:2">
       </div>
       <div class="column-one-half">
-        <img src="/public/images/examples/pm.jpg" alt="">
```
36degrees: Need to remove this file, or is it used elsewhere?

robinwhittleton: Might be used in prototypes so removing it would potentially be a breaking change. Might as well leave it in for the time being.

36degrees: 👍

selfthinker: If we know something should be removed at some point later, we should deprecate it. Do we have any official deprecation rules?

nickcolley:
Yay, we should include an alt tag so that assistive tech users know it's Larry 😻

PR #380, opened 2016-12-22
Add deploy key to let Travis push to GitHub

alexmuller:
I've generated an SSH keypair for this repository. The private key I encrypted using `travis encrypt-file` and committed to this repo as `.travis/govuk_elements.enc`. The public key is `.travis/govuk_elements.pub` which I added to the deploy keys for the repo - I gave it push access.

https://github.com/alphagov/govuk_elements/settings/keys

I've added a bit of code to the Travisfile to configure Git and SSH and a deployment which runs a script. I renamed the script from `jenkins.sh` (GOV.UK Jenkins used to run this script) and removed the conditional from the script because Travis handles that for us now.

This is better than using a `GH_TOKEN` because it's tied to a repo rather than a user. This means that:

- if the token is compromised because of a vulnerability in Travis, the attacker has access to all repos that the user can access, not just this repo
- if the user leaves GDS, the token will stop working

I'd recommend that the `update-release-branch.sh` script is changed to use this approach. Happy to help with that!


alexmuller:
I've deleted the Jenkins webhook from this repo's settings and disabled the job on GOV.UK Jenkins for this repo.

robinwhittleton:
Sat down with @alexmuller and ran through this. I’m happy with the approach, but it’s difficult to test without being merged. With that in mind let’s wait until we want to do a release in the new year, and merge this in at that point.

.travis.yml
```diff
@@ -2,6 +2,11 @@ language: node_js
 node_js:
  - "4.4.7"
 before_install:
+- git config --global user.name "Travis CI"
+- git config --global user.email "travis@travis-ci.org"
+- git remote add origin_ssh git@github.com:alphagov/govuk_elements.git
+- openssl aes-256-cbc -K $encrypted_85ebe8034b89_key -iv $encrypted_85ebe8034b89_iv -in .travis/govuk_elements.enc -out ~/.ssh/id_rsa -d
```
robinwhittleton: If this is generated by Travis CLI can we add a comment about this (and potentially how it can be regenerated in the future)?

.travis/govuk_elements.pub
```diff
@@ -0,0 +1 @@
+ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQCwxFgZUepuTRA3hW4JF4/u8zLEXaz9DaDUrsg1RoexiFUcHf5+BXIwjekM2wBj+0juBSVmWRqlRCX1ou0i6aTFk/9B8k+pgWw3sYoJNbNf3HFJXA/Ed2NsmBcKPQE9c5pMaJyWFZSg06BcuJ8WRNLM1QHCEJhTejR5Y6679iX4waml+0Gf22y4DDgsN/ikFmJRqzg1o3FdoyKfW4a3NTmsmI3dwQgMEq6wSlMZvXlyLYNSM6XXpbE9FqmzE3Zq2GR9R/l6q1P0b4ZWLvSp/v1o+oV30jQzuUbnBD9lC2a72P+1Zv/xe78FPDjIyoPn/lvL5WPXlYI/adMmiNmawlOpMHZxcG5kARUGLV+yCMSrIr8sOBJj7RhJB2dEQ8LzIv94IR64FVEQE6uaGGbGTeT0XjKDLnmB7OzTervlwhuus/ozxrX7WA/VCEfILjbAlSIG+yto0N8h/8izNAR1kWUBfY5/Q9SQhf8+pvwlB1f6qVk3BL5vGBeD6WajHlzoLVD85qZqkktliW8bJbAEE87BZ34Y7DEXRGUMHvDIhgJAcMu37KTnDvOdVcr1xxWkqrbgpyaFUMyQphFrjqIC0nalpnRD7ZWtv2QTlnCPYCOyifIyOE7bp2FMEhdMF8Eb976kF1y3BfvA0vthu7hgXxfxRn9Md+R24orhpBWSVvg5/Q== govuk_elements_deploy_travis
```
robinwhittleton: A readme in this folder as to what these files are doing would be useful.

alexmuller:
I sat down with @robinwhittleton, @gemmaleigh and @dsingleton yesterday and talked through this, and I've made a couple of small changes:

- moved git setup to `before_deploy` in Travis file
- added comment indicating that openssl command is automatically generated
- added README to Travis directory

I think this is good to go.

dsingleton:
👏  this is great - well documented and easy to understand, despite the complexity.

PR #387, opened 2017-01-17
Fix conditional match to use typeof

nickcolley:
We're missing a typeof in our conditional to check if something needs to be redirected.


gemmaleigh:
Thanks @nickcolley @colinrotherham @joelanman for spotting and fixing this.

PR #389, opened 2017-01-17
Remove the title from the cookie message link

robinwhittleton:
It’s unnecessary and doesn’t give an accessibility benefit. Thanks to @trevorsaint for catching this.

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -5,6 +5,6 @@
 module.exports = {
 
   // Cookie warning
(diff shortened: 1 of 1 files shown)


gemmaleigh:
Thanks @robinwhittleton, it was copied from the [config set for the prototype kit](https://github.com/alphagov/govuk_prototype_kit/blob/master/app/config.js), I'll update it there too.

Issue #327, opened 2016-10-17
Update minimum Node version to match LTS on 18/10


robinwhittleton:
Node is changing the LTS version tomorrow. We should change `package.json` to use this as the minimum supported version.

joelanman:
we'll also need to to test everything :)

joelanman:
it'll be v6

PR #393, opened 2017-01-21
Update advisory node requirement to 6.0

gemmaleigh:
6.0 is the LTS version, update package.json to use this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "repository": "alphagov/govuk_elements",
   "license": "MIT",
   "engines": {
-    "node": ">=4.4.7"
+    "node": ">=6.0"
   },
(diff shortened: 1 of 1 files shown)


robinwhittleton:
Fixes https://github.com/alphagov/govuk_elements/issues/327

robinwhittleton:
Why `<7.0`?

gemmaleigh:
To match the [govuk_prototype_kit](https://github.com/alphagov/govuk_prototype_kit/blob/71b6c5863285ede89237a088f3e164da6707c895/package.json#L7). 

I have removed it, it turns out I made this change in November without the `<7.0` and didn't open a PR.

PR #394, opened 2017-01-21
Update example of a single checkbox to use the new style

gemmaleigh:
This fixes #335.

#### What does it do?
- Rename the snippet
- Updates the example to use a custom checkbox
- Remove the text about large hit areas, as this is contradictory

#### Before:
before

#### After:
after

##### What type of change is it?

### DIFF
--- a/app/views/guide_form_elements.html
+++ b/app/views/guide_form_elements.html
@@ -234,21 +234,19 @@
 </code>
 </pre>
 
-  <h4 class="heading-small">Inline checkboxes</h4>
-  <ul class="list list-bullet text">
-    <li>large hit areas aren’t always appropriate</li>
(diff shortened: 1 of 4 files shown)


robinwhittleton:
I guess we want to be using a large checkbox. Was there any reason the old version was kept as a small one?

robinwhittleton:
There’s also [`.form-checkbox` and `.form-radio` styles left in the Sass](https://github.com/alphagov/govuk_elements/blob/f7cf0099ef47f1a7347b0ffe701231b37a77211c/public/sass/elements/_forms.scss#L172) that appear to be unused. We should remove this as part of this PR.

gemmaleigh:
@robinwhittleton I removed those additional styles, this should be good to go.

PR #395, opened 2017-01-23
add red border to erroneous date inputs

danboscaro:
Change to validation sass to target date pattern inputs so when the error class is added to the parent date pattern element the date inputs have a red border applied to them

There isnt an open issue, but if you look at the example on this page, the date pattern with an error is missing the red border - https://govuk-elements.herokuapp.com/form-elements/example-form-elements/ 

## What does it do?

Restores consistency with error styles

## How has this been tested?

Looking at the same page after the fix has been applied shows the date pattern input with a red border 

## What type of change is it?

### DIFF
--- a/public/sass/elements/forms/_form-validation.scss
+++ b/public/sass/elements/forms/_form-validation.scss
@@ -10,7 +10,8 @@
 
   // Form inputs should have a red border
   // Add a red border to .form-controls that are direct descendants
-  > .form-control {
+  > .form-control,
+  > fieldset > .form-date .form-control {
     border: 4px solid $error-colour;
   }
 


public/sass/elements/forms/_form-validation.scss
```diff
@@ -10,7 +10,8 @@
 
   // Form inputs should have a red border
   // Add a red border to .form-controls that are direct descendants
-  > .form-control {
+  > .form-control,
+  > fieldset > .form-date > .form-group > .form-control {
```
robinwhittleton: Can we change the new selector to just `.form-date .form-control`? Walking the tree doesn’t take any time in modern browsers, and it reduces the specificity hit.

danboscaro: It does need to be quite specific so not to trigger the border on fields that are in a panel where the parent question has an error on it eg...

- yes/no question that reveals a panel with a date field in when yes is clicked.
- a validation error is triggered on the yes/no question when its not selected.
- when the panel is revealed the date fields inherit the border if its `.error .form-date .form-control`

could be trimmed down to `> fieldset > .form-date .form-control`?

ive updated the pull request

gemmaleigh:
I'm not sure this is correct. This would ensure that all three fields have a red border - which would be consistent with other form elements, however, if one or more fields was filled in - then there would need to be a way to only apply a border to the field that was empty.

I think it would be better to apply a class to set the border to the form fields that haven't been filled in, rather than rely on the parent `.error` class to set this.

danboscaro:
Hi Gemma, thanks for the update

I understand what you are saying and having the ability to apply the border without depending on the parent error would be a good feature, however i don't want to muddy the waters of this issue

I see that the issue has originated from an update that was fixing another issue - https://github.com/alphagov/govuk_elements/commit/9c8d158b7ae2790f7666525a12103635f82812c7

The feature you're suggesting would be a breaking change as it'll require all current projects that use date fields to update their markup by adding classes to show the errors in the right way e.g http://govuk-frontend-alpha.herokuapp.com/components/browser/form-date-group--has-error

The fix in this pull request allows for current projects that were created before the change linked above to update their elements library without it affecting the way their date errors are shown

gemmaleigh:
Hi @danboscaro, 

The change I'm suggesting would involve renaming `.error`(where applied to a form group, to set a red left hand border) to `.form-group-error` and would also require any inputs that should have a red border - to use the new class `.form-control-error`. 

Since we've already had an issue related to this code - I'd prefer not to make it even more fragile by merging this PR. Instead I'm keen to have styles for errors and validation that don't require a parent element to set the styling for child elements.

If you need this fix, feel free to override it on your own project. 
I'm going to close this PR in favour of #405.

PR #397, opened 2017-01-30
Fix invisible input content when changing colours

selfthinker:
When overwriting colours via browser settings, Firefox and IE always default to black text in form fields, irrespective of said settings.
When the background of the form fields is transparent and the general background is dark (e.g. black), it results in black text on a black background, i.e. making the text invisible and therefore completely inaccessible.

It seems counterintuitive but due to browser bugs (which I will report) it is necessary to remove the `background-color: transparent`. The only downside I can see is that anyone wanting to use input fields on a non-white background would need to give it a background colour (which would actually re-introduce this bug again in Firefox).

A reduced test case to play around with: http://jsbin.com/sibube
See also alphagov/govuk_frontend_toolkit#377.

It makes text within form fields accessible to partially sighted users who change their browser's colour settings in IE or Firefox.

## What does it do?

It removes the transparent background from input fields.

## How has this been tested?

I tested this in
* Firefox and IE 11 and 8 with default colours, a dark and a light background via their browser settings
* Chrome with default colours, a dark and a light background via the extensions "High Contrast" and "Change Colors"
* Firefox, IE 11 and Edge with default colours, a dark and a light background via the Windows accessibility system settings
* Various browser with different colours via Mac OS, iOS and Android accessibility system settings

(Please ignore the buttons in those screenshots, I will open a different PR for that. Also ignore the wrong font.)

This is the page I was testing, showing the default colours:
firefox-default

### Before, i.e. showing the bug

Firefox:
firefox-dark-before

IE 11:
ie-dark-before

### After, i.e. with fix

Firefox:
firefox-dark-after

IE 11:
ie-dark-after

## What type of change is it?

### DIFF
--- a/public/sass/elements/_forms.scss
+++ b/public/sass/elements/_forms.scss
@@ -146,9 +146,9 @@ textarea {
   width: 100%;
 
   padding: 5px 4px 4px;
-  color: inherit;
-  background-color: transparent;
-  border: 2px solid;
+  // setting any background-color makes text invisible when changing colours to dark backgrounds in Firefox (https://bugzilla.mozilla.org/show_bug.cgi?id=1335476)
+  // as background-color and color need to always be set together, color should not be set either
+  border: 2px solid $text-colour;
 
   // TODO: Remove 50% width set for tablet and up
   // !! BREAKING CHANGE !!


robinwhittleton:
Yep, happy with this, with the addition of a link to the Firefox bug report in the commit message.

nickcolley:
In your example, if I set the `background` to black and the `color` to white, the text is not readable since it's the same color as the input. Is this a problem?

http://jsbin.com/bokikexizu/1/edit?html,css,js,output

selfthinker:
That only happens because I've set the input's colour to `inherit`. In a real world scenario you would also change the input's background and text colour. But as that is what triggers the bug, I cannot incorporate that into the example.
But your change just made me re-test my example without the `color: inherit`, just to make sure that has nothing to do with the bug. And it doesn't, so good to know. :)

nickcolley:
I must be missing something but we do currently use `color: inherit`?

https://github.com/alphagov/govuk_elements/pull/397/commits/fdb36eb84fa9eeeb1d9fbcf6427aa58244160809#diff-874ea218a8261b484fef9e13094e3b86R149

selfthinker:
Yes, we do. But that shouldn't have anything to do with the reduced test case. The test case shows the bug with or without `color: inherit`.

selfthinker:
I just chatted to @nickcolley separately. His concern is either services using different colours or users using user stylesheets. In both cases the fact that we have `color: inherit` would mean they would need to change something. I think that's an edge case. But we could remove `color: inherit` again, just to be sure. That would only mean I would need to re-test all the things again.

The form control colour was set to `inherit` quite recently in 735955e7ffb6198d2b5ad5348ed5eb941435646b. You could set the border colour via `border-color: inherit` as well instead.
The transparent background colour is quite recent as well (29f756b2c35a047ed74c4f83edae04460faf7760). I doubt it broke any colour settings from services or we would have heard from them? (Or maybe not, how often do we hear back in those cases?)

I personally don't think it's worth removing `color: inherit` but it probably wouldn't be harmful either. What do others think?

selfthinker:
I just filed a bug report for the issue in Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1335476
It probably doesn't make sense to report it for Internet Explorer as Microsoft only really updates Edge and that doesn't allow colour changes.

selfthinker:
I just updated the commit message to include the link to the bug report and I also extended the explanation a bit.

selfthinker:
Prompted by another unrelated bug report, I will change this PR to remove the `color: inherit` as well (@nickcolley). We had a [user report](https://govuk.zendesk.com/agent/tickets/1773705) that this means he sees white text on a white background. That we use `color: inherit` is one half of the problem (the other half is that the rendering app in question gives input fields a background colour).

selfthinker:
I have just updated the commit again accordingly. I have also added a comment into the SCSS file instead of just removing the offending lines as it's something that can easily be overlooked and potentially introduced again.

robinwhittleton:
OK, I’m happy with this. Thanks for all the hard work in getting this sorted.

PR #405, opened 2017-02-15
Breaking changes to form validation

gemmaleigh:

The styling for form validation is pretty fragile and relies on a parent `.error` selector to apply a red border to the left of a `.form-group` div. This class also sets a red border on any `.form-control` inputs that are direct descendents. 

Rather than relying on a child div sitting inside a parent for this to work - add two new classes `.form-group-error` and `.form-control-error`. One will set the border for form-groups and the other for inputs/textareas with a class of form-control.

This also solves the problem where there may be multiple `.form-control` elements within a parent `.error` - and only one of them needs to have a red border - an example of this is the date of birth pattern.

> This would be a breaking change, but at some point we'll need to make breaking changes, so let's have a discussion  about it here, rather than on Slack where it'll get lost.

I have added the error examples to the date of birth page:

d2e321d8-1ef7-42b3-5198-6b3114ddef96

### DIFF
--- a/app/views/examples/example_date.html
+++ b/app/views/examples/example_date.html
@@ -17,6 +17,83 @@
       <!-- Date of birth -->
       {% include "snippets/form_date.html" %}
 
+      {% include "snippets/form_date_show_errors.html" %}
+
+      <!-- Date of birth -->
+      <div class="form-group form-group-error">
+        <fieldset>
+          <legend>
+            <span class="form-label-bold">Date of birth</span>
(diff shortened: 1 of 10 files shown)


nickcolley:
Could we remove the nesting if we were to move towards this approach?

ztolley:
I've had to code around this issue for things previously but using an 'incomplete' class so happy for this change to happen, even if it means I have to revisit some code. Agreed that it should be a major version change to avoid breaking code

gemmaleigh:
@nickcolley's comments about nesting have been addressed, this is ready to merge @robinwhittleton.

Issue #375, opened 2016-12-08
Radio buttons not checking when outer circle only is clicked


charlottewild:
Case with two users who clicked the outer edge of the radio button, which activated the 'live' state of yellow but the radio button was not checked. User failed to notice. 

## How would you expect it to work?
Whole radio button should be clickable and instigate the radio button become active when clicked.

## How does it work currently?
User can click the outer ring (quite common now they are larger) but the radio button does not become checked.

## Feel free to suggest a fix...
Would suggest that should the outer ring be clicked, the radio button becomes checked without need for user to hit the middle of the target.

Caveat: this has only happened with two low-to-no users to date but thought I would share. Should it happen more, will update this ticket. 

## Explain the steps to reproduce the bug

1. Use of radio buttons (new large format) in any question on Gov.uk website

## Tell us about your environment

Radio buttons were being accessed in Google Chrome (unsure what version) on PC with Windows7 and being used on a GOV.uk prototype. No live example available.

robinwhittleton:
Thanks for the report! I’ve replicated something, but I want to check that it’s the same behaviour you’re seeing. I can get the focus ring to appear if I:

1. hold the mouse button down while hovering over the outer ring
2. mouse the cursor away from the radio
3. release the mouse button

This adds the focus ring to the radio but doesn’t select it. Is this the behaviour you were seeing in the lab?

charlottewild:
Thanks Robin. Yes, that sounds very similar. I have video footage that I can slice down to show the specific issue. Can I add video content to this thread to you know?

nickcolley:
That'd be really useful if you can manage it!

You could upload the clip to YouTube - as long as it does not have anything sensitive of course.

charlottewild:
Hi Nick, sorry for delay with this - was just wanting sign off from the Service Manager that she was happy for me to share. First clip is here - [https://youtu.be/0LCF37YLgOw](url) - am just editing the second down.

Issue #367, opened 2016-11-30
Provide a disabled style for radios and checkboxes


timpaul:
An interesting comment on our recent blog post asked if we'd created disabled styles for the new custom controls. I'm pretty sure we haven't. Whilst it's true that we want to discourage use of disabled form controls I wonder if we should at least have an appropriate style if they are used?

edwardhorsford:
We probably should. I thought about this last week and then promptly forgot about it.

edwardhorsford:
Half an idea. Would want to test heavily - I'm not at all sure it's clear.

Radio isn't contrast safe, but the text is.
![screen shot 2016-11-30 at 17 13 38](https://cloud.githubusercontent.com/assets/2204224/20763086/8c17b7f8-b720-11e6-9e88-9a547d34e360.png)

cjforms:
Hi all - hello! I'm back. 

What's the user need? 

It's possible - indeed, frequent - that there's an answer in the user's mind that isn't available in the official list of options. I've seen this implemented in two ways:
1. By not offering the answer
2. By offering the answer, and then implementing logic to explain why the answer isn't going to work for the service and to guide the user to an answer that's acceptable for the service.

Method 1 is easy but unhelpful for the user.
Method 2 counts as 'do the hard work to make it simple'.

What I've never yet seen is a disabled option. Not to say it's not out there somewhere on the Internet, but it's not in my extensive library of examples.

edwardhorsford:
Hi @cjforms,

We don't want to offer this - I'd expect our guidance would still remain "don't do it". 

With that said, it's an attribute that *can* be set on checkboxes and radios - we can't stop services using it. Right now if a checkbox or radio has the attribute set, they will be unclickable, but not otherwise visually different - this would look buggy and be really confusing to users.

Thankfully disabled radios and checkboxes are rare - great you've not seen one! There's an example [here](http://www.w3schools.com/jsref/tryit.asp?filename=tryjsref_radio_disabled2) (click the button to disable it)

---

Agreed on both of your alternate suggestions. If we write some guidance in the service manual on checkboxes and radios, I'd like to say something similar about alternatives to disabling fields.

cjforms:
Thanks for the explanation - and pleased to see that the example is a demonstrator, not something that's live.

Speculation: could the 'disabled' appearance change the thing from a target to the words 'don't do this'?

joelanman:
[...]

Issue #354, opened 2016-11-18
Googly eyes on radios


edwardhorsford:
Cat certain browser zooms, the new radio styles have the select state off centre.

I suspect it may happen with the checkboxes too, but it's harder to notice there.

## How would you expect it to work?
The selected state should be in the centre of the radios.

## How does it work currently?
At 75% and 125% the selected state is off centre.

75% example:
![screen shot 2016-11-18 at 11 35 22](https://cloud.githubusercontent.com/assets/2204224/20429266/7c0c7aa0-ad85-11e6-8e51-ff632c1fb139.png)
125% example:
![screen shot 2016-11-18 at 11 35 09](https://cloud.githubusercontent.com/assets/2204224/20429277/839b5afc-ad85-11e6-8363-eb29bc1b9570.png)

## Explain the steps to reproduce the bug
Open elements in Chrome, set browser zoom to 75% or 125%

## Tell us about your environment
Googly eyes seen on Mac OSx 10.11.6, in Chrome 54, Firefox 49 and Safari 10

robinwhittleton:
Chrome seems to be handling this better on Windows 10 at least now. No googly eyes at least, although there’s a single pixel white dot in the middle of the bullet at 75%.

36degrees:
Following the launch of the [GOV.UK Design System](https://design-system.service.gov.uk/), GOV.UK Elements will now only get major bug fixes and security patches, so I'm going to close this.

PR #406, opened 2017-02-16
Radios and checkboxes without JS

robinwhittleton:
This PR removes the [JavaScript requirement](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/javascripts/govuk/selection-buttons.js) for radio buttons and checkboxes. To do this the markup had been changed, but there are no visual changes for users (beyond IE < 9 having a focus box around just the input rather than the whole label).

The new radios and checkboxes still required the selection buttons JS. This is a potential point of failure that’s a hangover from the old grey-box implementation, and isn’t really needed. We originally went with it because removing it would require markup changes and would mean that it they weren’t a drop-in solution. I think that was probably a mistake though, so this corrects it.

While making the markup changes I took the opportunity to rename the component from the old `block-label` to the more descriptive `multiple-choice`. This has better developer semantics, but also has the benefit that it’ll obviously require markup changes.

This fixes https://github.com/alphagov/govuk_elements/issues/375 and https://github.com/alphagov/govuk_elements/issues/367.

Tested in all macOS native browsers, and in Browserstack for IE and Edge on Windows (back to 8) and a selection of iOS and Android versions. The only initial bug found was a problem in IE9/10 that was fixed. No assistive tech testing has been done yet, and I’d want to do at least some to make sure that no new issues have been introduced. The assistive tech bugs fixed during the development of the new radios and checkboxes shouldn’t have regressed as their implementation hasn’t changed.

### DIFF
--- a/public/sass/_govuk-elements.scss
+++ b/public/sass/_govuk-elements.scss
@@ -46,7 +46,7 @@
 @import "elements/details";                       // Details summary
 @import "elements/panels";                        // Panels with a left grey border
 @import "elements/forms";                         // Form - wrappers, inputs, labels
-@import "elements/forms/form-block-labels";       // Chunky labels for radios and checkboxes
+@import "elements/forms/form-multiple-choice";    // Chunky labels for radios and checkboxes
 @import "elements/forms/form-date";               // Date of birth pattern
 @import "elements/forms/form-validation";         // Errors and validation
 @import "elements/breadcrumbs";                   // Breadcrumbs
--- a/public/sass/_govuk-elements.scss
+++ b/public/sass/_govuk-elements.scss
@@ -46,7 +46,7 @@
 @import "elements/details";                       // Details summary
 @import "elements/panels";                        // Panels with a left grey border
 @import "elements/forms";                         // Form - wrappers, inputs, labels
-@import "elements/forms/form-multiple-choice";    // Chunky labels for radios and checkboxes
+@import "elements/forms/form-multiple-choice";    // Custom radio buttons and checkboxes
 @import "elements/forms/form-date";               // Date of birth pattern
 @import "elements/forms/form-validation";         // Errors and validation
 @import "elements/breadcrumbs";                   // Breadcrumbs
--- a/app/views/examples/example_form_elements.html
+++ b/app/views/examples/example_form_elements.html
@@ -40,7 +40,7 @@
                   This is the label text in bold
                 </span>
               </legend>
-              <div class="block-label">
+              <div class="multiple-choice">
                 <input id="example-radio-yes-no-yes" type="radio" name="example-yes-no-group" value="Yes">
                 <label for="example-radio-yes-no-yes">Yes</label>
               </div>
@@ -86,7 +86,7 @@
                   This is hint text
                 </span>
               </legend>
-              <div class="block-label">
+              <div class="multiple-choice">
                 <input id="radio-yes-no-yes-b" type="radio" name="yes-no-group-b" value="Yes">
                 <label for="radio-yes-no-yes-b">Yes</label>
               </div>
@@ -166,7 +166,7 @@
                   Error message goes here
                 </span>
               </legend>
-              <div class="block-label">
+              <div class="multiple-choice">
                 <input id="radio-yes-no-yes-c" type="radio" name="yes-no-group-c" value="Yes">
                 <label for="radio-yes-no-yes-c">Yes</label>
               </div>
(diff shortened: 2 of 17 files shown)


nickcolley:
Would this fix this issue? https://github.com/alphagov/govuk_elements/issues/354

public/sass/_govuk-elements.scss
```diff
@@ -46,7 +46,7 @@
 @import "elements/details";                       // Details summary
 @import "elements/panels";                        // Panels with a left grey border
 @import "elements/forms";                         // Form - wrappers, inputs, labels
-@import "elements/forms/form-block-labels";       // Chunky labels for radios and checkboxes
+@import "elements/forms/form-multiple-choice";    // Chunky labels for radios and checkboxes
```
gemmaleigh: The comment needs updating here. "Custom" radio buttons and checkboxes?

robinwhittleton:
No, that seems to be a generic browser rendering issue at different zoom levels (although it seems more pronounced in Chrome than other browsers). This doesn’t change the CSS rules that render the inputs (apart from fixing that IE bug).

nickcolley:
@robinwhittleton it seems as though, if we change the markup to add specific elements instead of pseudo elements we'd be able to fix that issue?

robinwhittleton:
Honestly I haven’t tried it, but I’m not sure it would make any difference. I’ll give it a shot though.

robinwhittleton:
I’ve tried out a few different ways of fixing https://github.com/alphagov/govuk_elements/issues/354 as part of this issue, but nothing is looking particularly promising. I’m loathe to complicate the markup here, especially as adding empty spans can cause issues for JAWS-on-IE and requires extra fragile workarounds. In the componentised future this wouldn’t be a problem, but where we’re relying on people cutting and pasting HTML I’m worried about the potential for accessibility problems in a small visual fix for a non-standard setup. It’s worth pointing out that the visual problem only manifests itself when the page is zoomed _out_, an unusual thing to do compared to zooming in.

tl;dr it’s worth leaving the issue open, but I don’t think we need to address it in this PR.

robinwhittleton:
Blocked by https://github.com/alphagov/govuk_elements/pull/394/files

nickcolley:
@robinwhittleton do you have a link to the the empty spans problem?

robinwhittleton:
It was found and fixed on an internal repo, but the gist is that with the markup:

```html
<label>
  <input type="radio" />
  <span><span></span></span>
  Example
</label>
```

The label text wouldn’t be read by JAWS-on-IE (JAWS-on-Firefox was fine). Some text needed to be added to the middle of the elements (we settled on `<span><span>&nbsp;</span></span>`) to get it read out. The old Verify custom radio builder shows this: https://github.com/alphagov/verify-frontend/blob/4f5b1592bc3e861dfd3f2294ddf9f816ceca01d3/lib/verify_form_builder.rb

nickcolley:
Very interesting, I wonder if that could be solved with `aria-hidden`?

robinwhittleton:
Potentially? In other news, I’ve rebased this and updated on top of [Gemma’s work to convert the joint text box + checkbox over to the new style](https://github.com/alphagov/govuk_elements/pull/394) so this is now unblocked for merging.

robinwhittleton:
…assuming the accessibility team are happy. @cfq / @accessiblewebuk / @selfthinker ?

selfthinker:
Should we give it another full testing round? Or should we only have a quick look? Or do you want us to test anything specific?

robinwhittleton:
Well, the whole markup has changed, but the CSS remains mostly the same. I guess we probably want to retest it in everything based on that?

selfthinker:
This is now at the top of our backlog, Richard will probably test these soon.
Do you have a URL for him to test?

robinwhittleton:
I’ve pushed the branch to http://govuk-elements-wip.herokuapp.com/form-elements/#form-radio-buttons . That’s a temporary site, but it’s only really @gemmaleigh and I who use it so we’ll let you know if we need to repurpose it.

accessiblewebuk:
I have completed a full round of accessibility testing (including WCAG 2.0 A and AA, JAWS14, JAWS17, NVDA 2017, Dragon 12.5 and 13, Read and Write Gold and Zoomtext) with the exception of VoiceOver on Mac which I can't do. No issues found.

robinwhittleton:
I integrated this into [a branch on my copy of Verify’s codebase](https://github.com/alphagov/verify-frontend/compare/master...robinwhittleton:new-new-custom-radios) (with a custom built gem). There wasn’t too much work in the end, it was mostly down to updating the markup in the tests.

robinwhittleton:
When this is merged and released we’ll want to deprecate selection-buttons.js in govuk_frontend_toolkit. See https://github.com/alphagov/govuk_frontend_toolkit/pull/389

gemmaleigh:
I'm happy to merge, this has been reviewed by the accessibility team - thanks @accessiblewebuk and @robinwhittleton!

PR #411, opened 2017-03-02
Remove service manual from header

36degrees:
- Elements is not really considered to be part of the service manual, but a separate entity.
- The service manual is no longer referred to as the 'Government Service Design Manual'
- The service standard is no longer referred to as the 'Digital by Default Service Standard'
- There is no link to the service toolkit
- There are extra unused template partials that can be removed

I made changes to the templates and viewed the changes in only Chrome.

##### Before
screen shot 2017-03-02 at 10 11 15

##### After
screen shot 2017-03-02 at 10 11 22

- Documentation change


edwardhorsford:
Not checked the code, but in principle this change makes a lot of sense. Thanks!

gemmaleigh:
Thanks @36degrees! 💯

PR #408, opened 2017-02-24
Make Travis deployment config consistent

dsingleton:
With other GDS FE repos, see:
https://github.com/alphagov/govuk_frontend_toolkit/pull/385

Steps specified in before_deploy are only run if a deployment is
triggered [1], which will only happen on master, which can't be
done by a user without write perms.

[1] [Travis before_deploy documentation](https://docs.travis-ci.com/user/deployment/npm/#Running-commands-before-and-after-deploy)

PR #414, opened 2017-03-07
Fix broken travis deployment setup

dsingleton:
This was broken in #408, where the setup was moved to `before_deploy`
to avoid permission issues when running PR tests in a way consistent
with other `alphagov` apps.

Previously that was guarded by a shell if, which was removed, but a
bit at the end was left and is breaking deployment builds. :(

- Fixes deployment builds.

Internal change, no effect on semver.


gemmaleigh:
👍  thanks for the fix!

PR #403, opened 2017-02-09
Update the Github templates for pull requests and issues

gemmaleigh:
#### What problem does the pull request solve?

It updates the pull request and issue templates, in an attempt to encourage their usage - and to provide structure to the issues raised and pull requests opened for GOV.UK elements.

#### What does it do?

- uses smaller headings 
- removes the checkboxes, as these imply that a choice is a set of things to be completed
- requests that lines that don't apply are deleted

#### What type of change is it?

- New feature (non-breaking change which adds functionality)

#### Put an `x` in all the boxes that apply
- I have read the **CONTRIBUTING** document.

PR #424, opened 2017-03-17
Ensure that table cells using tabular-numbers are 19px

gemmaleigh:

Fixes #403 - Tables have inconsistent front sizes.

Remove the core-16 mixin being applied to table cells with a class of
numeric. All tables should use 19px - to match the body text size.

Tested by running the GOV.UK elements app locally, see screenshots, or Heroku app preview.

#### Before:

before - tables

#### After: 

after - tables

### DIFF
--- a/public/sass/elements/_tables.scss
+++ b/public/sass/elements/_tables.scss
@@ -18,17 +18,17 @@ table {
 
   th {
     font-weight: 700;
-    // Right align headings for numeric content
-    &.numeric {
-      text-align: right;
-    }
+  }
+
+  // Right align table header cells and table cells with a numeric class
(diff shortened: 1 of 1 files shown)


robinwhittleton:
Yeah, I think this is an oversight. The `numerical` class was created in https://github.com/alphagov/govuk_elements/commit/da33ee5592f94cd59bf55edd87a54ddc04d64b39 at a time when all `td`s had `font-size: 16px`. That was removed from core `td`s in https://github.com/alphagov/govuk_elements/commit/c4352572664f8f51594d8cf660d1260590a598e4 but left for numeric ones.

PR #426, opened 2017-03-17
v3.0.1

gemmaleigh:
This release includes:

Update the govuk_frontend_toolkit to 5.1.2. This updates show-hide-content.js to work with new .multiple-choice custom radio buttons and checkboxes, released in GOV.UK elements 3.0.0.

A fix for table cells, these are now 19px when using tabular-numbers.


robinwhittleton:
https://github.com/alphagov/govuk_elements/blob/master/VERSION.txt was missed.

PR #429, opened 2017-03-21
Simplify the latest-release script

gemmaleigh:
Remove git config as this already exists (in the first script provider)

Force push the latest-release branch to origin_ssh (this is also used
by push-version-tag to update the tag).

### DIFF
--- a/update-release-branch.sh
+++ b/update-release-branch.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -ex
+set -e
 
 # Update the latest-release branch with the most recent tagged release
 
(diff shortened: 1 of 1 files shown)


dsingleton:
👍 

Happy to merge this, but I think we could simplify things further in the future:

- The `if` statement in `update-release-branch.sh` tests it's not a PR and is `master` - this should now be implicit from the deploy provider config.
- `update-release-branch.sh` and `push-version-tag.sh` are becoming pretty similar and could be merged to reduce duplication
- The combined code from those two scripts could be merged directly in to `create_release.sh`, so that everything is in on place and easier to understand (eg `origin_ssh` being set and used in the same script).

gemmaleigh:
I've made those changes over in #431.

PR #434, opened 2017-03-24
Page header philsherry

philsherry:
Change the page header structure

The current design pattern seems like it could be built a little better.
The page title reads "Check your answers — GOV.UK elements",
which isn't how the `h1` reads. Having a `span` in the `h1` gives
differing results, depending on the assistive technology used.

 • VoiceOver reads:
    "Heading level onetwo items. GOV.UK elements. Check your answers."
    ("onetwo" read as one word)
 • NVDA reads:
    "Heading level one. GOV.UK elements. Heading level one. Check your answers."
 • JAWS reads:
    "GOV.UK elements. Check your answers. Heading level one."

This new version splits the page name (`h1`) and the section name (`p`) –
which are now in the correct source order – and then positions the section
name above with CSS. This keeps the pattern visually identical, but brings
added accessibility wins.

The user can skip to the first header and they'll hear *just* the page title.

 • VoiceOver reads:
    "Heading level one. Check your answers."
 • NVDA reads:
    "Heading level one. Check your answers."
 • JAWS reads:
    "Heading level one. Check your answers."

The next element reads out the section name, with some visually-hidden content.

 • VoiceOver reads:
    "This section is [section name]"
 • NVDA reads:
    "This section is [section name]"
 • JAWS reads:
    "This section is [section name]"


gemmaleigh:
@selfthinker can you please advise on these changes to the structure of the headings for govuk-elements?

gemmaleigh:
@philsherry Thanks for your PR, sorry it's taken so long for me to get to it. I've asked Anika on our accessibility team to review.

selfthinker:
Sorry, I've just been looking through some old issues and found my name here. Either I hadn't seen it at first or forgotten about it.
I will be on leave next week and will likely not have time to look at this before then. But I will put it on my todo list and look at it when I get back.

adamliptrot-oc:
@selfthinker have you had chance to review this?

selfthinker:
Sorry, I had looked into this a couple of months ago but delayed responding and then forgot about it.

Back then I also discussed with @edwardhorsford and we agreed that having this type of page header which has essentially two types of headers in one doesn't make sense in the context of GOV.UK Elements. It doesn't make sense to repeat "GOV.UK elements" above every h1.
That makes it quite difficult to review.

It would make sense on an example page to show how to style it and when to use it. But even then how to mark it up depends on the context. Sometimes what you proposed would make sense, other times it wouldn't.
It wouldn't make sense for the most common type where something above the main heading tells you which type of document a page is, for example https://www.gov.uk/guidance/grenfell-tower-fire-june-2017-support-for-people-affected

I wonder if it makes sense to a) remove "GOV.UK elements" everywhere and b) create a section under Typography which shows this component in different contexts.

Do we know which other contexts we have across government?

nickcolley:

We have taken this same problem into consideration in [some updates to the GOV.UK Design System](https://github.com/alphagov/govuk-design-system/pull/461).

Thanks for raising, sorry it has taken so long to reply.

PR #435, opened 2017-03-27
Resolve 'dead area' between radio/checkbox and label that has hand cursor but is not clickable

andymantell:
On the new radio buttons / checkboxes, there is a 12px gap between the input control and the label which has a hand cursor (inherited from the `.multiple-choice` parent div) but is not clickable. Since it has the hand cursor, I feel it _should_ be clickable.

The alternative would be to not show the hand cursor in this 12px dead area but that doesn't feel right to me?

Checked in Chrome (Ubuntu), IE8, IE9, IE10, IE11, Edge 14, iOS9 Safari, Android 6 Chrome.
Can test more if necessary, if this change is approved. Just let me know...

image

### DIFF
--- a/public/sass/elements/forms/_form-multiple-choice.scss
+++ b/public/sass/elements/forms/_form-multiple-choice.scss
@@ -8,20 +8,11 @@
   clear: left;
   position: relative;
 
-  padding: (8px $gutter-one-third 9px 50px);
+  padding: 0 0 0 38px;
   margin-bottom: $gutter-one-third;
-
-  cursor: pointer; // Encourage clicking on the label
-
-   // remove 300ms pause on mobile
-  -ms-touch-action: manipulation;
-  touch-action: manipulation;
 
   @include media(tablet) {
     float: left;
-    padding-top: 7px;
-    padding-bottom: 7px;
-    // width: 25%; - Test : check that text within labels will wrap
   }
 
   // Absolutely position inputs within label, to allow text to wrap
@@ -43,6 +34,18 @@
 
   label {
     cursor: pointer;
+    padding: 8px $gutter-one-third 9px 12px;
+    display: block;
+
+    // remove 300ms pause on mobile
+    -ms-touch-action: manipulation;
+    touch-action: manipulation;
+
+    @include media(tablet) {
+      float: left;
+      padding-top: 7px;
+      padding-bottom: 7px;
+    }
   }
 
   [type=radio] + label::before {


nickcolley:
Nice spot thanks for the PR!

I've attached some screenshots where the pink is anywhere `cursor: pointer` is enabled, and where the current clickable area is before and after your proposal.

|Current|Proposed|
|-|-|
|screen shot 2017-03-27 at 15 53 00screen shot 2017-03-27 at 15 52 41|screen shot 2017-03-27 at 15 53 22screen shot 2017-03-27 at 15 53 27|

andymantell:
Good point, I'll have a peep

andymantell:
@nickcolley If I just remove the `cursor: pointer` from the parent div, the label and input inside it already have a pointer on them, so you end up with what you highlighted in red.

Only thing I wonder is whether to add the tiniest amount of top and bottom padding onto the label to give it some height. What do you reckon?

edwardhorsford:
Nice catch!

It's important that radios and checkboxes have large hit areas. In the examples above ideally the entire pink area would be clickable - as the hand suggests.

andymantell:
@edwardhorsford Ah I see sorry, I interpreted the pictures wrong. You reckon actually the entire of the light pink area should be clickable, in both those examples? What about this example:

!(https://cloud.githubusercontent.com/assets/2445413/24362846/ec2e663c-1305-11e7-9b61-7d4dc6120de3.png)

Should it be clickable all the way to the far right hand side there?

andymantell:
Happy to do that - I just wondered whether the right hand edge there was quite a long way away from the end of the word Ireland. Is it strange for it to be clickable that far away, or should it be more like this:

image

nickcolley:
Willing to be vetoed but I'd vote for it being all the way on smaller screens.

Thinking about where people's thumbs will be placed if they miss on the right side of the screen we'll catch their input if it's full width.

This will be much more evident in small labels also like the examples above these 'Yes', 'No'

andymantell:
This is a bit more complicated now we don't have the grey box round them, to give that visual indication that there's a _thing_ there. That's my only concern - that a large blank area of the screen as a clickable thing might be more "surprising" than it is useful.

joelanman:
is there a downside to wrapping the whole thing in `label` like we used to? That means you have a nice big hit area

nickcolley:
@joelanman the label is being used with pseudo elements for the fake input, if we wrapped the label we'd most likely need another element alongside the input that is nested.

@robinwhittleton was worried that extra elements will play worse with screenreaders

andymantell:
Don't think we need to change the markup. We should be able to do either of the hit area options with what we've got. Just need to decide on what the "right" hit area is

On 27 Mar 2017 5:02 p.m., Joe Lanman <notifications@github.com> wrote:

is there a downside to wrapping the whole thing in label like we used to? That means you have a nice big hit area

andymantell:
What do people think about the attached screenshots, with the clickable areas being fatter, but not extending full width? (See the dotted red borders around checkboxes and radios)

The alternative suggested above is that on the mobile option, the clickable area extends full width.

image

image

nickcolley:
I think unless someone else has strong objections it may be better to keep the full width behaviour as is but with your fixes to touch areas and then raise an issue to look into this again later?

andymantell:
:+1: Cool, will sort that out

andymantell:
Ok that will leave us with the following click areas (dotted in red). I'll tidy up let you know when it's ready for review

image

image

edwardhorsford:
Looking good!

andymantell:
Cool, cheers @edwardhorsford. I think this is ready for review in that case.

nickcolley:
Will give it another look over thanks so much @andymantell this is great.

nickcolley:
Tested this on a Galaxy S7, iPhone 6, IE 6, IE 8, IE 9 and Chrome.

LGTM 👍 

Will leave open for a little bit since these are important components for others to comment then merge.

public/sass/elements/forms/_form-multiple-choice.scss
```diff
@@ -43,6 +39,14 @@
 
   label {
     cursor: pointer;
+    padding: (8px $gutter-one-third 9px 12px);
```
robinwhittleton: Do we need the parentheses here?

public/sass/elements/forms/_form-multiple-choice.scss
```diff
@@ -8,19 +8,15 @@
...
-
    // remove 300ms pause on mobile
   -ms-touch-action: manipulation;
   touch-action: manipulation;
 
   @include media(tablet) {
     float: left;
-    padding-top: 7px;
-    padding-bottom: 7px;
     // width: 25%; - Test : check that text within labels will wrap
```
robinwhittleton: I wonder if we can just dump this comment with the other two lines. It’s been [hanging around since 2014](https://github.com/alphagov/govuk_elements/commit/7835339e070063d566e98df2162b725cc59c379d).

public/sass/elements/forms/_form-multiple-choice.scss
```diff
@@ -8,19 +8,15 @@
...
 
-  padding: (8px $gutter-one-third 9px 50px);
+  padding: 0 0 0 38px;
   margin-bottom: $gutter-one-third;
 
-  cursor: pointer; // Encourage clicking on the label
-
    // remove 300ms pause on mobile
   -ms-touch-action: manipulation;
   touch-action: manipulation;
```
robinwhittleton: The `touch-action` lines: do we want to move them into the `label` section as well? Ideally we’d minimise the code these are applied to, as (for example) they prevent double-tap zoom. (thanks to @nickcolley for pointing that out)

robinwhittleton: Although if we want to test and make that change as a follow-up I’m happy with that too.

public/sass/elements/forms/_form-multiple-choice.scss
andymantell: If it's been around since 2014 it probably relates to a way earlier version anyway, and is probably irrelevant?

public/sass/elements/forms/_form-multiple-choice.scss
andymantell: Something in my head is saying it's something to do with an old sass compiler which got upset without them. But that may be imagined nonsense. And if it were ever true, it's not true any more, it compiles fine without them so I'd support removing them yeah.

public/sass/elements/forms/_form-multiple-choice.scss
andymantell: Seems like a fair point to me, and worth including rather than spinning it out into a separate issue. The interaction _has_ entirely moved off the div and onto the label.

public/sass/elements/forms/_form-multiple-choice.scss
robinwhittleton: Yep, drop it.

public/sass/elements/forms/_form-multiple-choice.scss
robinwhittleton: Good stuff, let’s drop them.

public/sass/elements/forms/_form-multiple-choice.scss
andymantell: Done

public/sass/elements/forms/_form-multiple-choice.scss
andymantell: Done

public/sass/elements/forms/_form-multiple-choice.scss
andymantell: Done

robinwhittleton:
Left open for a day with no extra comments, and @nickcolley’s testing looks good so let’s merge.

nickcolley:
Thanks a lot @andymantell 💯  🚀

36degrees:
Good stuff! Thanks! 👏

andymantell:
No probs, thanks for everyones help

Issue #130, opened 2015-11-30
Details polyfill doesn't open/close with spacebar/enter key when JAWS is running


gavboulton:
In user testing we've seen a partially sighted user unable to open the details element when using JAWS screen reader. The summary is described as a button, when pressing the spacebar/enter key, the key name is read out but the element never opens.

JAWS seems to prevent the spacebar/enter key from being captured, unless another key is pressed before. So, if you press the Windows key and then spacebar the details element does open.

I've been able to replicate this problem using Vista/IE9/JAWS 13. I've tried adding `aria-pressed` attributes to the summary element and have tried replacing `role="group"` with `role="application"` but neither of these have worked.

edwardhorsford:
@aduggin you might want to take a look at this?

robinwhittleton:
This seems to be specific to JAWS on IE.

selfthinker:
This also came up during our testing of Elements. @accessiblewebuk found that although JAWS and NVDA work fine in other browsers, both cannot access the details in IE 11 (and probably older IEs). It is keyboard-accessible when JAWS and NVDA are switched off, but not when one of them is switched on.

I started looking at this today and found the reason. Some screen readers map the screen reader key + space (or enter) to the **click** event. As you can see in [Event Handlers and Screen Readers](http://unobfuscated.blogspot.co.uk/2013/05/event-handlers-and-screen-readers.html), the click event is usually the **only** event fired for IE and JAWS/NVDA combinations.

Our polyfill is currently listening for the keyup, keypress and mouseup events. It used to listen to the click instead of the mouseup event, but that was [changed because it made it trigger twice in Chrome](https://github.com/alphagov/govuk_elements/commit/66ae6ce45a75a74fd4dcc61ecac8ad432a93b6b1). That's also the reason why I didn't simply create a PR as it doesn't seem to be that easy to know which events are the right ones.

I've had a quick look at some other accessible details polyfills:
* [better-details-polyfill](https://github.com/chemerisuk/better-details-polyfill) uses click and keydown
* [jquery-details](https://github.com/mathiasbynens/jquery-details) uses click and keyup

Now I hope to wiretap the hive mind here. (I'm not too experienced with JS events.) There must be some universal fix for this, right? Which events should trigger the opening of details?

If you'd like to play around with various events, I have [isolated all the important bits](http://jsbin.com/vucicis).

selfthinker:
[...]

PR #440, opened 2017-04-04
Fix details polyfill for JAWS and NVDA in IE

selfthinker:

As reported in #130, the details/summary cannot be opened when using JAWS or NVDA in Internet Explorer.
As part of my investigation I also found that using the enter key on the summary would also submit a form in IE if details is used within a form.
This fixes both issues.

Some screen readers map the screen reader key + space (or enter) to the **click** event. As you can see in [Event Handlers and Screen Readers](http://unobfuscated.blogspot.co.uk/2013/05/event-handlers-and-screen-readers.html), the click event is usually the **only** event fired for IE and JAWS/NVDA combinations.

Our polyfill is currently listening for the keyup, keypress and mouseup events. It used to listen to the click instead of the mouseup event, but that was [changed because it made it trigger twice in Chrome](https://github.com/alphagov/govuk_elements/commit/66ae6ce45a75a74fd4dcc61ecac8ad432a93b6b1).

I've now changed it to use click and keydown. As far as I know, it does not trigger twice anymore.
I've also done some minor refactoring.

I have not tested within Elements but have [isolated all the important bits into a JSbin](http://jsbin.com/cerajed). (I have added some debugging under that, not in a `console` because I wanted to easily test in older and mobile browsers.) See the [current version with the bugs and the debugging in a JSbin](http://jsbin.com/neheyey) for a comparison.

I have tested in various browsers and screen reader combinations, mainly IE and Firefox with JAWS and NVDA, both and Chrome on its own. I found that it's more likely to be troublesome if the browser doesn't [support `details`](http://caniuse.com/#feat=details), so I concentrated on testing in older browsers via Browserstack (Firefox 48 and under, Safari 5.1 and under, Android Browser 3 and under, Windows Phone; there is no version of Chrome which is old enough). I tested with mouse and keyboard and touch and Dragon when I could.

Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/public/javascripts/govuk/details.polyfill.js
+++ b/public/javascripts/govuk/details.polyfill.js
@@ -10,6 +10,8 @@
   'use strict'
 
   var NATIVE_DETAILS = typeof document.createElement('details').open === 'boolean'
+  var KEY_ENTER = 13
+  var KEY_SPACE = 32
 
   // Add event construct for modern browsers or IE
   // which fires the callback with a pre-converted target reference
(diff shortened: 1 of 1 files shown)


selfthinker:
I can see that @gemmaleigh has now deployed this to Heroku. (Thanks! Or does that happen automatically?)
That means you can now test this change in-situ on https://govuk-elements-review-pr-440.herokuapp.com/typography/#typography-hidden-text

gavboulton:
❤️

gemmaleigh:
@selfthinker I set it up so it happens automatically for each PR, I thought it might be useful for others to help out with reviewing.

Thanks so much for your work on this! 💯

gemmaleigh:
There's only one thing I have noticed. 

In Chrome, the current behaviour is that you can press either space, or enter to toggle the hidden content. 

With this solution, only using the spacebar works.

From this article: http://accessibleculture.org/articles/2012/03/screen-readers-and-details-summary/ 

_Clicking or pressing the Enter or Space keys on the summary element should toggle the visibility of the details element's content._

So I think we'd want to preserve that existing behaviour.

selfthinker:
Thanks @gemmaleigh, I didn't spot that. I tested in Chrome, but this shows that it's better to test in the real environment because it works in my JSBin but not when used within Elements.
I'll have a look.

selfthinker:
I have pushed a new version.
It fixes previously mentioned bugs plus a new bug we found: The space key would not open the details in Firefox (it would trigger two clicks, so it would open and close quickly again).
Sorry, I usually like granular commits but I tried and couldn't disentangle everything, so it is all in one single commit unfortunately.

Fixing the enter key was easy but that created a new bug and fixing that led to even more bugs... So, all in all it turned out to be more complex than I initially though. A big thanks to @tvararu, @joelanman and @splitbrain who helped me solve most issues.

I re-tested the automatically deployed version in Firefox and IE 11 in Windows 10 with and without JAWS and NVDA, macOS Firefox, Chrome, Safari (all latest), Edge and Chrome in Win 10, Firefox 45 (a version which doesn't support details), IE 9, IE 8, IE 7, Safari 5.1 and some more.
This version doesn't work as expected in two of those browser combinations:

* IE 8 and under don't work with the enter key (but work with space and click)
* Safari 5.1 and under only work with click, so neither space nor enter works; I suspect the same Safari version in iOS wouldn't work either when a keyboard is attached, but I don't know how to test that

I haven't tried to debug those two new issues yet. I will do that tomorrow.
I would think the IE 8 issue is the least important because we only support IE 8-10 on a functional level and it is functional enough. The Safari 5.1 issue bugs me though, as it would mean no keyboard access at all. If it's too difficult to fix, I might leave it as it is.

selfthinker:
I have fixed IE 8 and under, and Safari 5.1 now works with enter but not with space. I'm happy with that.
All in all, this fixes things in 3 modern browser combinations (IE with JAWS or NVDA, IE without them and Firefox) and partly breaks something in one old browser (Safari 5.1 and under).

gemmaleigh:
Thanks @selfthinker! Very happy to merge this. 👍

PR #443, opened 2017-04-06
Fix poor alignment of text in radios example

### LINKED ISSUE #442: Radio text badly aligned
The last radio example on [this page](http://govuk-elements.herokuapp.com/form-elements/example-radios-checkboxes/) had badly formatted text.

#### How would you expect it to work?
The content about the properties should be beneath the bold of the option

#### How does it work currently?
Content badly flows.

#### Feel free to suggest a fix...
I think the copy should probably be part of the label, not outside of it.

Screenshot:
screen shot 2017-04-06 at 17 19 04

tvararu:
Closes #442.

Has the added effect of providing more information to users of assistive technology, which seems desirable. I tested in VoiceOver and the behaviour is that the bold part of the label gets pronounced, followed by a short pause, followed by the rest of it, which is ok.

I have not looked to see if this problem exists in other places. [Digging here might uncover more places](https://github.com/alphagov/govuk_elements/commit/4fb688e208fb65a04d5ccde59700ee4d73c39b7d).

## Before

screen shot 2017-04-06 at 18 21 28

## After

screen shot 2017-04-06 at 18 21 20

### DIFF
--- a/app/views/examples/example_radios_checkboxes.html
+++ b/app/views/examples/example_radios_checkboxes.html
@@ -138,14 +138,18 @@
 
             <div class="multiple-choice">
               <input id="radio-part-2" type="radio" name="housing-act" value="Part 2">
-              <label for="radio-part-2"><span class="heading-small">Part 2 of the Housing Act 2004</span><br></label>
-              For properties that are 3 or more stories high and occupied by 5 or more people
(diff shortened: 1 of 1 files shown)


gemmaleigh:
Thanks @tvararu!

PR #451, opened 2017-04-12
Increase the size of the text in the highlight box

### LINKED ISSUE #200: Highlight box colour contrast too low for WCAG AA

gemmaleigh:
Increase the heading size by one level and use large text (36px) for
the text which is currently failing contrast checks.

For smaller screens, the font-size (of the large text - as [we're using the govuk_frontend_toolkit's core-36 mixin](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/_typography.scss#L73)) will be 24px.

Discussion over in #200.

#### Before:

screenshot-localhost-3000-2017-04-12-16-14-13

#### After:

after

small-screen

- Bug fix (non-breaking change which fixes an issue) Fixes #200.

### DIFF
--- a/app/views/guide_colour.html
+++ b/app/views/guide_colour.html
@@ -371,12 +371,12 @@
         </h1>
 
         <div class="govuk-box-highlight">
-          <h2 class="bold-large">
+          <h2 class="heading-xlarge">
             Application complete
           </h2>
-          <p>
+          <p class="font-large">
             Your reference number is <br>
-            <strong class="heading-medium">HDJ2123F</strong>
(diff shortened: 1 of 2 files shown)


markhurrell:
👍🏻 looks great

PR #454, opened 2017-04-26
Corrected numbered list rules

keithiopia:
Numbered lists should be written as sentences (start with a capital, end with a full stop).
This is confirmed in https://www.gov.uk/guidance/how-to-publish-on-gov-uk/markdown#numbered-list 
and
http://govspeak-guide.herokuapp.com/#lists-of-steps

### DIFF
--- a/app/views/guide_typography.html
+++ b/app/views/guide_typography.html
@@ -133,9 +133,11 @@
 </pre>
 
   <h3 class="heading-medium" id="typography-lists">Lists</h3>
-  <p class="text">
-    List items start with a lowercase letter and have no full stop at the end.
-  </p>
+  
+  <ul class="list list-bullet text">
+    <li>list items start with a lowercase letter and have no full stop at the end</li> 
+    <li>numbered list items start with an uppercase letter and do have a full stop on the end</li>
+  </ul>
 
 <div class="example">
   {% include "snippets/typography_lists.html" %}
--- a/app/views/guide_typography.html
+++ b/app/views/guide_typography.html
@@ -136,7 +136,7 @@
   
   <ul class="list list-bullet text">
     <li>list items start with a lowercase letter and have no full stop at the end</li> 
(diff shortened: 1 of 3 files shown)


36degrees:
Is this a duplicate of #238?

keithiopia:
@36degrees I wasn't aware of that other one.
It doesn't sound like they actually resolved that discussion. As it is we have a piece of guidance that contradicts all our other published guidance. 
This caused a problem in a department where someone followed our style guide when writing a numbered list, then their dev changed it all because they were following the elements.

app/views/guide_typography.html
```diff
@@ -133,9 +133,11 @@ <h3 class="heading-medium" id="typography-links">Links</h3>
 </pre>
 
   <h3 class="heading-medium" id="typography-lists">Lists</h3>
-  <p class="text">
-    List items start with a lowercase letter and have no full stop at the end.
-  </p>
+  
+  <ul class="list list-bullet text">
+    <li>list items start with a lowercase letter and have no full stop at the end</li>
```
36degrees: Given the conversation in the other PR, can I suggest rather than duplicating we just defer to the style guide here, like we do in the markdown documentation?

e.g.

> See the [style guide](https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#bullet-points-steps) for guidance on how to punctuate lists.

We'd still update the examples.

keithiopia: Yeah that works for me, thanks @36degrees

36degrees: Are you OK to update the PR?

keithiopia: Sure, So leave my snippet changes as they are, but replace the stuff in the main view with a link to the style guide?

gemmaleigh:
Thanks @keithiopia and @36degrees, I've updated the example typography page too.

PR #456, opened 2017-04-27
Windows10 select box contrast ratio in IE and Edge

igloosi:
Accessibility fix on Windows 10 IE and Edge select box active 
Issue : https://trello.com/c/LGKKreic/461-tech-spike-insufficient-colour-contrast-of-select-box-items

Tested on IE11 and Edge 15 where the issue existed.
Also tested on latest Chrome. Firefox and Safari to make sure nothing was affected there. 

#### Before:
before

#### After:
after

### DIFF
--- a/public/sass/elements/_forms.scss
+++ b/public/sass/elements/_forms.scss
@@ -9,6 +9,7 @@
 // 4. Form hints
 // 5. Form controls
 // 6. Form control widths
+// 7. Browser accessibility fixes
 
 // 1. Helpers
 // ==========================================================================
@@ -222,4 +223,14 @@ textarea.form-control {
   @include media(tablet) {
     width: 12.5%;
   }
+}
+
+// 7. Browser accessibility fixes
+// ==========================================================================
+
+option:active,
+option:checked,
+select:focus::-ms-value {
+  color: $white;
+  background-color: $govuk-blue;
 }


public/sass/elements/_forms.scss
```diff
@@ -223,3 +224,13 @@ textarea.form-control {
     width: 12.5%;
   }
 }
+
+// 7. Browser accessibility fixes
+// ==========================================================================
+
+option:active,
```
selfthinker: Are you sure `:active` is needed? My gut feeling said `:checked` is weird but from testing all 3 of them just now, it seems like `:checked` is needed for the activate state when the select box is open and `:focus::-ms-value` is needed for the active state when the select box is closed (but still focussed). Do you know for what case `:active` is needed?

selfthinker:
I also just tested if the adding the colours does anything when changing colours via browser or system settings. (We found earlier that that can sometimes be challenging.) But that all looks good.

I also just noticed that the colour combination of white and our govuk-blue fails WCAG AAA (both variations of foreground and background). But I assume that must be a known issue and that we intentionally didn't fix it (for our links). (I think I vaguely remember something about it being difficult to find a blue that is not too dark so that it's difficult to distinguish from normal text around it.)

edwardhorsford:
`govuk_blue` is not intended to meet WCAG AAA, so I think it's fine. Still - it's _frustratingly_ close to being compliant...

public/sass/elements/_forms.scss
igloosi: to retain the correct background colour as you press down with a mouse

PR #459, opened 2017-05-05
Replace Grunt with Gulp.js

gemmaleigh:
This PR removes Grunt as a dependency and instead adds [Gulp](http://gulpjs.com/).

The GOV.UK Frontend Alpha uses Gulp, the GOV.UK Prototype kit uses Gulp and there has been work on both of those projects that it would be useful to include in this project. That is only possible when GOV.UK elements also uses Gulp, hence this change.

The GOV.UK Frontend Alpha focused on the delivery mechanism (using Gulp) for building a package of frontend assets. This will be useful to this repo - and will help with the separation of GOV.UK elements (the app and the online guide) from the govuk-elements-sass npm package. 

The main changes are: 

Assets are moved from /public to the /assets directory and instead, the /public directory is generated when the app is run.

Heroku will now only run server.js, rather than the Grunt - copy, sass and watch tasks. 

In gulpfile.js the following tasks are added:

Clean
Deletes the /public and /dist directories

Styles
Compiles CSS from Sass
Outputs both a minified and non-minified version into /public/stylesheets/

Images
Copies images to /public/images

Scripts
Copies JavaScript to /public/javascripts

Build
Runs tasks that copy assets to the public directory

Package 
This task is in preparation for another PR to separate the govuk-elements-sass files from the app, by copying these files to /dist

Server
Runs the app and configures nodemon

Test
Adds mocha and tests that the the assets are copied to /public and that the app runs

Watch
When a file is changed, re-run the build task.

Develop
Runs the build task to copy assets, sets up watches and starts the app.

Default
Lists the three main tasks (gulp, develop, package).

- I have updated the the documentation.


fofr:
Could you include a reason why the switch from Grunt to Gulp.js is being made?

PR #461, opened 2017-05-08
Remove `aria-describedby` from erroring form fields and hint text

selfthinker:
The `aria-describedby` attribute was added before we moved the error message or hint text for a field into its label (or legend in case of radio/checkboxes) in 87449f013521d09ca2b51df113f4c85c85ec36d8.

Now that it's also in the label/legend, it gets read out multiple times to screen readers. Therefore it is not necessary anymore and should be removed.

The error message and/or hint text is read out twice to screen readers. Example pages where that happens:
* http://govuk-elements.herokuapp.com/errors/example-form-validation-multiple-questions
* http://govuk-elements.herokuapp.com/form-elements/example-date/

I tested with JAWS in IE 11, NVDA in Firefox and VoiceOver in Safari.
They all behave the same when tabbing to the input field. Before they were reading the additional text twice and with this change only once. When using the down arrow key, only JAWS read the additional text twice, the others didn't.

### DIFF
--- a/app/views/examples/example_date.html
+++ b/app/views/examples/example_date.html
@@ -24,13 +24,13 @@
         <fieldset>
           <legend>
             <span class="form-label-bold">Date of birth</span>
-            <span class="form-hint" id="example-dob-hint-1">For example, 31 3 1980</span>
+            <span class="form-hint">For example, 31 3 1980</span>
             <span class="error-message">Error message goes here</span>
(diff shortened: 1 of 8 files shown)


gemmaleigh:
Thanks @selfthinker, I have tested locally with VoiceOver in Safari and it works for me 👍 , happy to merge.

PR #462, opened 2017-05-10
Build the govuk elements sass package

gemmaleigh:
This PR aims to separate the govuk-elements-sass package from the GOV.UK elements app. 

**How it works now:**

The GOV.UK elements app shares its package.json file with the govuk-elements-sass npm package.

Confusingly, the package.json file has the name govuk-elements-sass, but its dependencies are those needed for the app, not for the package of sass files.

**How it should work:**

The GOV.UK elements app has its own package.json file, the dependencies listed are those to run app.

When a new version is released:

The GOV.UK elements app builds the govuk-elements-sass package to `/dist` and the contents of `dist` (a versioned .tgz file) is then released to npm.

The versioned .tgz file is uploaded to GitHub releases.

**This PR:**

Creates the govuk-elements-sass package by:
* copying the sass files to a temporary folder
* creating a new package.json file and copying it to the temporary folder (this file is unique to the package, rather than sharing the apps dependencies)
* building a package of the sass files for npm - copying over the README, CONTRIBUTING, LICENCE and VERSION files. 
* updating the npm ignore file to ignore all files in this repo, excluding the dist directory.

> This will be a breaking change as the govuk-elements-sass package currently has a folder structure which includes `/public/sass`. This PR will remove those two parent directories, so the package now contains:

```
CHANGELOG.md
LICENSE.txt
README.md
_govuk-elements.scss
elements/*
node_modules/govuk_frontend_toolkit/
package.json
```

The contents of `/dist` will be released using [GitHub Releases using Travis](https://docs.travis-ci.com/user/deployment/releases/). 

> The next release of govuk-elements-sass will include this breaking change

- Yes - references to `/public/sass` have been removed.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,5 +1,2 @@
 /*
-/public/*
-/public/sass/*
-!/public/sass/_govuk-elements.scss
-!/public/sass/elements
+!/dist/*
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,7 @@ deploy:
   api_key:
     secure: VAHPK3W0YVWqx03j659FbKjvEwd+UIDfb1x55to9u+PzyVqQp1CXalsYuKQSavxqInaV3gzNJpbANo6SIXlpAUxZPChiTk0rJwUhz7u1fRY0Lshb+aDOMnXSfjSqiIGThNRpjYdxBiJsnrYP1T6unHuo2A1K8mXjwdFtxhadh+w=
   email: govuk-dev@digital.cabinet-office.gov.uk
+  skip_cleanup: true  # Don't cleanup /dist
   on:
     tags: true
     all_branches: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,14 @@ deploy:
   on:
     tags: true
     all_branches: true
+- provider: releases  # For tagged commits, upload the contents of /dist to GitHub releases
+  api_key:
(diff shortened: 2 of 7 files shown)


tvararu:
From a high level perspective this gets the job done, and there's less `gulp`ing around than I expected there to be when I initially read the description. 👍 

That being said the `buildNpmPackageJson` task feels wrong and kind of obfuscates the process of creating that package. The fact that the package is created as a build artifact as opposed to being part of VC can make it harder to test and get a hold of.

An approach I've seen other projects adopting for multi-package repos is using [lerna](https://www.npmjs.com/package/lerna). I have no experience with it but it seems like another way of solving this issue.

I'm not sure how much time there is to yak shave this for or how important it is. The approach outlined seems to work, albeit in a non-standard way that we need to take ownership of and maintain. So IMO we can ship this. But if using lerna would accomplish the same but with less config / using a well recognised tool that could be better.

gemmaleigh:
Thanks @tvararu, the only downside I can see to lerna is that `lerna publish` has to be run manually, for each release.

At present, Travis will release to npm when the VERSION.txt file is updated. 

I'll open a PR using the lerna approach and we can discuss, I think if we had more than one package - it would be really useful (for releasing components as individual npm packages), but I'm not sure we'll get the benefit with a single package.

tvararu:
@gemmaleigh I don't know how complex lerna is but you might be totally right that for one package it's not worth the investment.

I think this can go out as is. 👌  (re-read the code, gave it green)

Only thing possibly worth doing extra is writing out the body of this PR as some form of Markdown document for future reference. (so we don't need to dig up the PR if someone asks how this works)

gemmaleigh:
Having thought a bit more about this, I'm going to close this PR in favour of not doing this in a non-standard way. It's much simpler to keep the package files in version control, than generate them as part of a 'package' step - so I'll do that instead.

PR #478, opened 2017-06-01
Move breadcrumb outside the main element

### LINKED ISSUE #343: Skiplink moves focus to before the phase banner

gemmaleigh:

This allows users to skip directly to the page content, avoiding the repetition of the breadcrumb on each page.

The `<main>` tag is wrapped with a containing div `site-wrapper`, to establish a central column.
The breadcrumb is a child of the site-wrapper container. 

The `<main>` element starts after the breadcrumb. 

GOV.UK currently has the breadcrumb inside the `<main>` element, ideally, the two should be consistent.

Fixes #343.

### DIFF
--- a/app/views/guide_alpha_beta.html
+++ b/app/views/guide_alpha_beta.html
@@ -2,11 +2,9 @@
 
 {% block page_title %}Alpha and beta banners — GOV.UK elements{% endblock %}
 
-{% block content %}
+{% block main_content %}
 
 <main id="content" role="main" tabindex="-1">
-
-  {% include "includes/breadcrumb.html" %}
 
   <h1 class="heading-xlarge">
     <span class="heading-secondary">GOV.UK elements</span>
--- a/app/views/guide_buttons.html
+++ b/app/views/guide_buttons.html
@@ -2,11 +2,9 @@
 
 {% block page_title %}Buttons — GOV.UK elements{% endblock %}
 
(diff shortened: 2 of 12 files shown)


gemmaleigh:
@tombye can you please confirm this fixes your issue. 

If so, we'll need to let GOV.UK Publishing know to amend their breadcrumbs too.

cc. @nickcolley.

tombye:
Yep, this fixes it and provides a nice description of the changes needed if people follow this approach. 👍

nickcolley:
I've added this to our support board: https://trello.com/c/IsxWIQXQ/105-investigate-work-needed-to-have-breadcrumbs-consistent-with-govuk-elements

Issue #400, opened 2017-02-08
Details polyfill not available in any of the gov repos when npm install'ed


andymantell:
If you `npm install` the gov template, the toolkit and elements (govuk-elements-sass), the expectation is that you should have everything you need to use the gov styles. However, the <detail> element polyfill is missing since govuk-elements-sass only includes the sass.

Currently the polyfill is here in elements: 

https://github.com/alphagov/govuk_elements/blob/master/public/javascripts/govuk/details.polyfill.js

I wonder whether it should perhaps live here instead:

https://github.com/alphagov/govuk_frontend_toolkit/tree/master/javascripts

In my situation, I would then be able to `import` the polyfill along with any other bits of toolkit JS I require.

The alternative option might be to publish the polyfill as it's own repository. If it's a good thing that has wider utility beyond gov, perhaps we could consider releasing it in it's own right. I'm not sure of the provenance of the code though - it looks like the original may have come from a blog post? That may complicate matters?

_Another_ alternative might be to cast the net wider and use an open source details polyfill. If we could find one that was 95% good, and open PRs against it for any accessibility issues etc then that might be a good option rather than maintaining our own.

robinwhittleton:
@gemmaleigh can you remember why this was the only bit of JS not moved from elements to toolkit? I’ve forgotten the context.

gemmaleigh:
Hi @andymantell,

The details summary polyfill hasn't been moved to the govuk frontend toolkit, as it doesn't have tests to accompany it. There is also an issue that has been raised with it not working as expected when using JAWS with IE. 

There is work being carried out by the accessibility team to evaluate the patterns in GOV.UK elements, before we move the polyfill to the govuk frontend toolkit - I'd like to be sure this is the recommended pattern for toggling hidden content.

@cfq - can you let us know if you have found any issues with the details summary pattern please?

andymantell:
Ah, thanks @gemmaleigh. I'll commit it into our application's repository for the time being.

RobinKnipe:
@andymantell or @gemmaleigh - any news on if/when this might be resolved - or what I might use instead? Thanks.

gemmaleigh:
Hi @RobinKnipe apologies for the slow reply, we will be moving the details polyfill to the GOV.UK Frontend Toolkit.

Will update this issue once it is available via the [govuk_frontend_toolkit](https://github.com/alphagov/govuk_frontend_toolkit).

RobinKnipe:
Thanks!

PR #479, opened 2017-06-01
Use govuk_elements as a Node.js dependency

RobinKnipe:
# Added a version number to `package.json'.

Allows the project to be installed as a dependency with NPM.

Tested by inclusion into another project.
This project can now be added as a dependency in any Node.js project as described in the following steps:
 - install the dependency
```bash
npm install --save alphagov/govuk_elements
```
 - require the code
```javascript
require('govuk-elements/assets/javascripts/govuk/details.polyfill');
```

## About the change:

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,22 @@ GOV.UK elements uses [standardjs](http://standardjs.com/), an opinionated JavaSc
 Both linters run on CI to ensure that new pull requests are in line with them.
 
 
+## Add govuk_elements to your project
+
+### Install with NPM
+From the command line, run the following to include elements as a dependency in your project:
+```bash
+npm install --save alphagov/govuk_elements
+```
+
+### Require the code
+Code can be required in the standard way (even on the client if using `browserify`):
+```javascript
+// polyfill `details` elements on older browsers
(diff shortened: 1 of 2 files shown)


gemmaleigh:
Hi @RobinKnipe, 

Thanks for this. If you'd like to use GOV.UK elements - the sass files are available as an NPM package. 

https://github.com/alphagov/govuk_elements/tree/master/packages/govuk-elements-sass

The govuk_elements application isn't intended to be included as a dependency in this way, that's why the package.json file in the root isn't versioned.

I'm not sure it's a good idea to require the entire app as a dependency to include a single file.

I'll raise an issue to move the details polyfill to the [govuk_frontend_toolkit](https://github.com/alphagov/govuk_frontend_toolkit/tree/master/javascripts), where the rest of our shared JavaScript lives.

gemmaleigh:
We have an issue already: https://github.com/alphagov/govuk_elements/issues/400.

Sorry there hasn't been any movement on it, will try to prioritise it!

RobinKnipe:
Hi @gemmaleigh - thanks for getting back to me. I'm happy to get the polyfill from the govuk_frontend_toolkit if that's easier. I'll close this PR...

PR #482, opened 2017-06-02
Fix table spacing

selfthinker:
Remove right padding from last table cells.
I also removed the `color` declaration as it isn't necessary.

This is a design-y problem. Can a designer please fill this section in and explain what the problem is? @edwardhorsford raised it.

Ed: it makes things look neater, and matches the padding (or lack thereof) used on the other side.

### Before
spacing-before

### After
spacing-after

## What type of change is it?

I'm not 100% sure if it's a breaking change? If it is, it's only breaking things minimally.

### DIFF
--- a/assets/sass/elements/_tables.scss
+++ b/assets/sass/elements/_tables.scss
@@ -20,6 +20,11 @@ table {
     font-weight: 700;
   }
 
+  td:last-child,
+  th:last-child {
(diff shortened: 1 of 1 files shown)


gemmaleigh:
Thanks @selfthinker & @edwardhorsford 👍

PR #483, opened 2017-06-02
Only make column headers bold, not row headers

selfthinker:

Having (row) headers is very important for accessibility.
But all of our table headers are bold, and that often doesn't look good with long row headers.
People who understand table semantics would need to overwrite that.
And people who don't understand semantics might choose not to use them.

We often have badly formatted tables. One reason for that is how govspeak works (I will raise an issue for that later). I suspect that another reason is that when you format a table correctly, it often doesn't look good. (But I don't have any evidence to back that up.)

E.g. if you formatted the tables on our [bank holiday page](https://www.gov.uk/bank-holidays) correctly, it would currently look like this:
bank-holidays
(Although that doesn't use Elements but the example would be the same with Elements.)
And ideally, those headers shouldn't be bold. Although, there should be a way to make them bold (class for cells or modifier class for tables?).

This changes that so that only `th`s within a `thead` are bold. That might be controversial (and also be a breaking change), so I'd like your opinion on this. The question is also, how many services are using row headers?
What do you think?

This is how the table on the bank holiday page looks now, i.e. how a table within Elements would look with this change if it had row headers:
bank-holidays-original

## What type of change is it?
Although, I'm not 100% sure if it's really a breaking change? If someone's row headers turn non-bold because of this, would that be "breaking" it?

### DIFF
--- a/assets/sass/elements/_tables.scss
+++ b/assets/sass/elements/_tables.scss
@@ -16,7 +16,7 @@ table {
     border-bottom: 1px solid $border-colour;
   }
 
-  th {
+  thead th {
     font-weight: 700;
   }
 


edwardhorsford:
This feels like a good change. Having all row headers bold is distracting and likely not what we want in most cases. It makes sense that authors might avoid using them because of this styling. Non-bold feels like a sensible default.

joelanman:
Agree with non bold row headers

edwardhorsford:
@selfthinker can you include a screenshot of what the proposed change looks like?

selfthinker:
@edwardhorsford, if you can point me to a table using row headers, I could. But there is none within Elements. Not sure if it makes sense looking for an example used by a service (because I'm sure they would overwrite it)?

edwardhorsford:
@selfthinker what about the one above? Just to give an example of what it would look like. (and good if we ever refer to this PR again).

selfthinker:
Which one above? The table on the bank holiday page? That one doesn't have any row headers, I only used it as an example how it would look like if it **had** row headers. If it had them and it was using Elements and this was merged, it would look exactly the same as it looks now.

edwardhorsford:
@selfthinker exactly - I'm suggesting including a screenshot of here for posterity. In the future it may not look the same - better to have it as part of the PR.

selfthinker:
@edwardhorsford, okay, I've updated the original description with an additional screenshot.

PR #484, opened 2017-06-02
Move form help text into legend

selfthinker:
This is an alternative solution to #485.

Most groups of checkboxes and radio buttons within Elements have an additional introductory help text which isn't a hint. It's usually there to help less technical people understand that you can select multiple checkboxes, but can also be used for other purposes.

This moves that text into the legend to also make it accessible to assistive technologies like screen readers. The styling is nearly the same as before and adds a class to the current paragraph declaration to re-use its styling.

There is a bit more spacing above the help text in the new version (probably `0.26316em` which translates to 5px with default settings), even though it uses the same CSS. That is because previously the margins were collapsing as the `p` came directly after the `h3` and now they aren't anymore because the text has been moved into the legend. I'm not sure if it's worth fixing as spacing within Elements is quite inconsistent and it doesn't look wrong.
Let me know if you think I should do something about it.

## How has this been tested?

I've tested in multiple browsers via BrowserStack.

### Before
form-help-text-before

### After
form-help-text-after

### Differences
form-help-text-differences

## What type of change is it?

### DIFF
--- a/assets/sass/elements/_forms.scss
+++ b/assets/sass/elements/_forms.scss
@@ -20,6 +20,11 @@ fieldset {
   width: 100%;
 }
 
+// Hack to let legends or elements within legends have margins in webkit browsers
+legend {
+  overflow: hidden;
+}
+
 // Fix left hand gap in IE7 and below
 @include ie-lte(7) {
   legend {
--- a/app/views/examples/example_radios_checkboxes.html
+++ b/app/views/examples/example_radios_checkboxes.html
@@ -31,12 +31,13 @@
           What is your nationality?
         </h1>
 
-        <p>If you have dual nationality, select all options that are relevant to you.</p>
-
         <div class="form-group">
           <fieldset>
 
-            <legend class="visually-hidden">What is your nationality?</legend>
+            <legend>
+              <span class="visually-hidden">What is your nationality?</span>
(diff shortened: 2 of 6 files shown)


assets/sass/elements/_forms.scss
```diff
@@ -20,6 +20,11 @@ fieldset {
   width: 100%;
 }
 
+// Hack to let legends or elements within legends have margins in webkit browsers
```
selfthinker: This is quite hacky but is the least hacky solution I could find for fixing an issue in all webkits which will not allow a margin on the form hint text. I suspect this might be too controversial (a similar solution was tried in #76 by @timpaul and rejected). That's why I created an alternative solution to fix the same problem with #485.

igloosi: could use  `overflow: hidden;` instead

selfthinker: Ah, yes that works as well. I could have sworn I tried it but apparently not. I think that makes more sense, so will change it.

selfthinker:
I have added some documentation (as suggested by @gemmaleigh on #485), changed the "hack" to use `overflow:hidden` instead (as suggested by @igloosi) and re-tested in a couple of browsers.

gemmaleigh:
This looks good, thanks @selfthinker and @igloosi 👍

PR #485, opened 2017-06-05
Associate form help text with form

selfthinker:
This is an alternative solution to #484.

Most groups of checkboxes and radio buttons within Elements have an additional introductory help text which isn't a hint. It's usually there to help less technical people understand that you can select multiple checkboxes, but can also be used for other purposes.

This associates that text with the form via `aria-describedby` to make it more accessible to assistive technologies like screen readers. Ideally this should be put into the legend instead, but that proves to be difficult (see #484).

## How has this been tested?

I tested in the most used screen reader and browser combination: VoiceOver in Safari on macOS, JAWS in IE 11 on Windows 10 and NVDA in Firefox on Windows 10. They don't necessarily say the help text under all circumstances but always under some circumstances (e.g. when down-arrowing vs tabbing).

## What type of change is it?

I'm not quite sure if this needs to be documented or not. We don't officially have the concept of a "form help text". (That's the name I gave it, that might need improving.) But I am aware of forms in the wild which use something similar. Where would a documentation fit? Maybe under "hint text" as that is related.

### DIFF
--- a/app/views/examples/example_radios_checkboxes.html
+++ b/app/views/examples/example_radios_checkboxes.html
@@ -31,10 +31,10 @@
           What is your nationality?
         </h1>
 
-        <p>If you have dual nationality, select all options that are relevant to you.</p>
+        <p id="help-text-nationality">If you have dual nationality, select all options that are relevant to you.</p>
 
         <div class="form-group">
-          <fieldset>
+          <fieldset aria-describedby="help-text-nationality">
 
             <legend class="visually-hidden">What is your nationality?</legend>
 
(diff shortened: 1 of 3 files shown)


gemmaleigh:
@selfthinker which of these solutions, #484 or #485 is your preferred choice?

gemmaleigh:
We have a section on fieldsets and legends. 
http://govuk-elements.herokuapp.com/form-elements/#form-fieldsets

The guidance for moving this help text into the legend, or using `aria-describedby` to associate the two can go there.

selfthinker:
I'd prefer solution #484. As Richard Morton keeps on saying: "The [first rule of ARIA](https://www.w3.org/TR/using-aria/#rule1) is: Don't use ARIA."
#484 solves it without the help of ARIA and without the need for developers to take care of adding additional things that can go wrong (`id` and referencing it correctly to `aria-describedby`).

Let me know if the "hack" in #484 is okay with you, then I can add the documentation to it. Otherwise I will add it in here instead.

gemmaleigh:
Let's go with #484. Fine to include the "hack" since it'll make the spacing underneath hint text consistent. Thanks Anika.

gemmaleigh:
Closing in favour of #484.

PR #488, opened 2017-06-06
Add guidance on using table captions

selfthinker:

This is adding guidance on using table captions in addition to the work in #476 as suggested by @gemmaleigh.
I'm not sure how much should go into it. If we wanted to link to anything useful, I'd link to https://www.w3.org/WAI/tutorials/tables/caption-summary/


gemmaleigh:
This is perfect, thanks!

PR #489, opened 2017-06-07
Allow for a single-file import of all elements stylesheets

lennym:
Move the import of the govuk-elements styles into a single, standalone file so they can be imported as a single import statement without also including govuk_frontent_toolkit files.

We use a sass compiler which is npm-aware (https://npmjs.com/npm-sass), and so generally don't use configured `includePaths` when compiling. However, the main sass file in this repo is only compilable if the `includePaths` are set.

By moving the govuk-elements-sass code (as distinct from the govuk_frontend_toolkit code), we can `@import "govuk-elements-sass"` in our stylesheets and have them compile correctly (note the `sass` property added to package.json supports this shorthand, as opposed to having to specify a full file path)

This will also support any users who wish to import the elements code alone, without also importing govuk_frontend_toolkit for whatever reason.

### DIFF
--- a/assets/sass/_elements.scss
+++ b/assets/sass/_elements.scss
@@ -1,0 +1,31 @@
+// GOV.UK elements
+
+@import "elements/helpers";                       // Helper functions and classes
+
+// Generic (normalize/reset.css)
+@import "elements/reset";
+
+// Base (unclassed HTML elements)
+// These are predefined by govuk_template
+// If you're not using govuk_template, uncomment the line below.
+// @import "elements/govuk-template-base";        // Base styles set by GOV.UK template
+
+// Objects (unstyled design patterns)
+@import "elements/layout";                        // Main content container. Grid layout - rows and column widths
+
+// Components (chunks of UI)
+@import "elements/elements-typography";           // Typography
+@import "elements/buttons";                       // Buttons
+@import "elements/icons";                         // Icons - numbered steps, calendar, search
+@import "elements/lists";                         // Lists - numbered, bulleted
+@import "elements/tables";                        // Tables - regular, numeric
+@import "elements/details";                       // Details summary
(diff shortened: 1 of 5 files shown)


gemmaleigh:
Thanks @lennym!

The relationship between the govuk-elements-sass files and the govuk_frontend_toolkit isn't explicit (ideally the imports would be at the top of each file - not that it would resolve the issue with includePaths - but at least the relationship between the two would be clearer.)

I've been trying to avoid making breaking changes - which is why they're still listed at the top.

This looks good, I'm hoping to release a new version of govuk-elements-sass shortly and will include this in that release.

lennym:
Thanks a lot @gemmaleigh.
One thing I might suggest,

gemmaleigh:
I won't be able to merge this PR as-is, as the package is build from assets in `assets/`, which is where the files will need splitting. Are you happy to do this?

lennym:
I think I see what you mean... the same set of changes should be made to the files in the `assets` directory in the repo root?

Will do. Give me 5 minutes.

gemmaleigh:
You'll only need to make the changes once in the assets directory, there's a `gulp package` task which I'll run before publishing the package, that'll copy everything to /packages/.

https://github.com/alphagov/govuk_elements/blob/master/gulpfile.js#L127

lennym:
I've pushed the changes in assets - actually an extra change to also group the govuk-frontend-toolkit files as well, as IMO this makes it a lot clearer what is coming from where.

If you want me to undo the changes in `packages` then I'll `rebase -i` them out.

gemmaleigh:
If you could undo the changes in packages (aside from the change to the package.json file) and ensure those two new files are copied over, that would be great - thanks!

lennym:
Done. Let me know if I need to do anything else.

Issue #320, opened 2016-09-20
Clarification regarding using `<h1>` as field descriptors


joefitter:
Guidelines for [labels](http://govuk-elements.herokuapp.com/form-elements/#form-labels) states that all form fields should be given labels, the second point states that these can be hidden if the surrounding context makes them unnecessary.

Due to these points we have inconsistencies throughout our forms and the ways labels, legends and `<h1>`s represent fields.

On some (multi-step) forms/pages the field label is displayed as the page `<h1>` with the field directly below. In this case it seems like elements suggests adding a label/legend and hiding it, as per the example [here](http://govuk-elements.herokuapp.com/form-elements/#form-radio-buttons).

My concerns regarding this approach are firstly visual users will only see the title, but screenreader users will have the title read out, and then the identical `visuallyhidden` label read out afterwards. Secondly we have some steps which contain more than one field/label, so by using this pattern of header-for-label, the first field will appear to have more weight over the others.

For my first point I wanted to check if having a label/legend and an identical h1 is necessary? Could we use aria-describedby or aria-labelledby (or something else) to link the field directly with the H1 without the need for a label? Another suggestion would be to lose the H1 and use CSS to display the label/legend with H1 styling? If semantically a page requires a H1 and a label for a field, how come we are hiding them for visual users but not screenreader users?

For my point regarding multiple fields on a page, we have gotten around this on certain forms by using the `<h1>` as a section heading usually spanning a few pages (contact details, enquiry details etc) and in addition always displaying a relevant label/legend alongside fields and fieldsets.

Any clarification/guidance would by greatly appreciated as we seem to be coming up against this issue each time a form is developed!

edwardhorsford:
Hi Joel,

This has come up several times on the accessibility mailing list (are you on it?). I don't think we have a solution to your first problem yet. My understanding is that pages require an h1, and fieldsets require a legend. By hiding the legend, we get visual simplicity for sighted users. Sadly this means it's repeated for AT users. We'll hopefully look at this shortly, but I don't know what the solution will be.

To your second question - where there is more than one question / field on the page, it makes sense for the `h1` to be more of a section heading, with labels and legends as appropriate for the various questions.

gemmaleigh:
Fixed by #507, where the recommended solution is to move the `<h1>` inside the `<legend>` element - to avoid this duplication, where there is one question per page the heading and legend text are identical.

PR #507, opened 2017-06-28
De-duplicate legends/headings

selfthinker:

When we have one thing per page, our current pattern is to let the h1 be the question and then repeat but visually hide it in the legend (or label). That means screen readers will read the question twice, which is not a proper barrier but annoying and gets raised as an issue again and again, e.g. in #320 or [on the cross-government accessibility mailinglist](https://groups.google.com/a/digital.cabinet-office.gov.uk/forum/#!searchin/accessibility-community/error$20summary$20heading%7Csort:relevance/accessibility-community/oekFWa1LhdE/n2-pwcTHAQAJ).

This changes that by **moving the h1 into the legend**. That creates the following issues:

* It is currently invalid. But it will be valid in [HTML 5.2](https://www.w3.org/TR/html52/sec-forms.html#the-legend-element) and the W3C validator passes it fine. (See the [reason for the change](https://github.com/w3c/html/issues/724).) As such it might cause issues with existing validators run during CI. Does anyone know if this will be a problem for them?
* Because the h1 is moved further down into the code inside other elements, some systems/frameworks might make it difficult to do. Does anyone already know they would have problems with that? We could add some guidance that using the old way would be fine in that case (i.e. visually hiding the legend).
* When the label is in an h1 (see below), then also having the hint in there could potentially be too much. Although that will more be a problem for examples that use potentially unnecessary long hints. And please note, that is not a problem for the legend, only for the label as another element for the hint can exist alongside the h1.

## What else could solve this?

I tested some other ways of fixing this issue, some of which have been suggested by the cross-government accessibility community in the thread mentioned above:

* Visually hide the heading instead of the label: It doesn't solve the problem
* Remove the legend and use `aria-labelledby` instead: It solves the problem (except JAWS and legend combination) but feels wrong, 7 out of 10 testing tools fail this, although all tested AT is fine with it
* Remove heading and use `role=heading` instead: It solves the problem but feels even more wrong, the h1 is too essential to lose (even if it's there for AT, it would be missing for other things, e.g. search engines), 4 out of 10 testing tools fail this, although all tested AT is fine with it

### Browsers
I tested this in various old and new browsers via BrowserStack and it's not causing any issues anywhere. I have also built a local version with the question being a label for a text input and the label being in an h1. That didn't cause an issue either.
Neither has had any visual effect in any of the tested browsers.

### Screen readers

I tested in a couple of screen readers:
JAWS with IE11, NVDA in Firefox and VoiceOver on iOS read everything once (instead of twice) when "down-arrowing". VoiceOver on macOS reads everything twice (instead of three times) and reads the heading after it announced the group.

### Accessibility testing tools

I also wanted to make sure that accessibility testing tools wouldn't report any unnecessary issues, so tested this in aXe, Tenon, AChecker, WAVE, FAE, EIII, HTML_CodeSniffer, SortSite, Google A11y Dev Tools and Asqatasun. And only EIII came back with one fail for the legend ("The name of the form control is not set correctly"). As I mentioned above, all the other viable solutions failed way more times than this solution.

### How to test yourself

If you want to test this solution yourself, I'd suggest to not use this branch because it only includes one single type of implementation, i.e. a single question for radio buttons with a legend. I have tested all the difference scenarios I could think of, including when the question is for a text input, i.e. it would need to be in a label not in a legend. Although I have tested that locally within a version of the prototype kit with GOV.UK branding and Elements styles etc, I have created three separate unstyled JSbin version, so you can test just the markup independently:

* [Heading in legend with radios](http://jsbin.com/jexekiv)
* [Heading in legend with inputs](http://jsbin.com/powavec)
* [Label in heading](http://jsbin.com/xecoli)

## What is potentially missing? / Open questions

* [ ] I wonder if we should add more examples? Although there are lots of examples, they are only of one type of implementation, the other two types are missing.
* [ ] Do we need to add to the documentation that falling back to visually hiding the legend is acceptable in case someone has problems implementing this solution?

## What type of change is it?

### DIFF
--- a/app/views/examples/example_radios_checkboxes.html
+++ b/app/views/examples/example_radios_checkboxes.html
@@ -27,15 +27,13 @@
 
       <form action="/" method="post" class="form">
 
-        <h1 class="heading-large">
-          What is your nationality?
-        </h1>
-
         <div class="form-group">
           <fieldset>
 
             <legend>
-              <span class="visually-hidden">What is your nationality?</span>
+              <h1 class="heading-large">
+                What is your nationality?
+              </h1>
               <span class="body-text">If you have dual nationality, select all options that are relevant to you.</span>
             </legend>
 
@@ -64,15 +62,13 @@
           </fieldset>
         </div>
 
-        <h1 class="heading-large">
-          Where should we send your new<br> passport?
-        </h1>
-
         <div class="form-group">
           <fieldset>
 
-            <legend class="visually-hidden">
-              Where should we send your new passport?
+            <legend>
+              <h1 class="heading-large">
+                Where should we send your new passport?
+              </h1>
             </legend>
 
             <div class="multiple-choice">
@@ -87,14 +83,12 @@
 
             <div class="panel panel-border-narrow js-hidden" id="example-other-address">
 
-              <h2 class="heading-medium">
-                Your other address
-              </h2>
-
               <fieldset>
 
-                <legend class="visually-hidden">
-                  Your other address
+                <legend>
+                  <h2 class="heading-medium">
+                    Your other address
+                  </h2>
                 </legend>
 
                 <div class="form-group form-group-compound">
@@ -123,15 +117,13 @@
           </fieldset>
         </div>
 
-        <h1 class="heading-large">
-          Which part of the Housing Act was your licence issued under?
-        </h1>
-
         <div class="form-group">
           <fieldset>
 
             <legend>
-              <span class="visually-hidden">Which part of the Housing Act was your licence issued under?</span>
+              <h1 class="heading-large">
+                Which part of the Housing Act was your licence issued under?
+              </h1>
               <span class="body-text">Select one of the options below.</span>
             </legend>
 
--- a/app/views/examples/example_form_elements.html
+++ b/app/views/examples/example_form_elements.html
@@ -561,7 +561,7 @@
 
       <div class="form-section">
         <h3 class="heading-medium">
-          These examples are one-question-per-page examples and use a heading for the question, with a visually hidden legend inside a fieldset.
+          These examples are one-question-per-page examples and use a heading for the question inside the legend of the fieldset.
         </h3>
         <!-- "Chunky" Radio buttons, stacked -->
         {% include "snippets/form_radio_buttons.html" %}
--- a/app/views/snippets/form_checkboxes.html
+++ b/app/views/snippets/form_checkboxes.html
@@ -1,10 +1,8 @@
-<h3 class="heading-medium">Which types of waste do you transport regularly?</h3>
-
 <form>
   <fieldset>
 
     <legend>
-      <span class="visually-hidden">Which types of waste do you transport regularly?</span>
+      <h3 class="heading-medium">Which types of waste do you transport regularly?</h3>
       <span class="body-text">Select all that apply</span>
     </legend>
 
--- a/app/views/snippets/form_inset_checkboxes.html
+++ b/app/views/snippets/form_inset_checkboxes.html
@@ -1,13 +1,11 @@
-<h1 class="heading-medium">
-  What is your nationality?
-</h1>
-
 <form>
   <div class="form-group">
     <fieldset>
 
       <legend>
-        <span class="visually-hidden">What is your nationality?</span>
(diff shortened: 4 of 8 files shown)


app/views/examples/example_radios_checkboxes.html
```diff
@@ -64,15 +62,13 @@ <h1 class="heading-large">
...
-        </h1>
-
         <div class="form-group">
           <fieldset>
 
-            <legend class="visually-hidden">
-              Where should we send your new passport?
+            <legend>
+              <h1 class="heading-large">
+                Where should we send your new passport?
```
fofr: Does this still need the line-break that was in the `h1`?

fofr:
Presumably in a circumstance where the H1 and the legend text differ we'd still advocate having both. Not sure what would justify the difference though.

This is a really nice PR that answers all of the questions I had about this pattern. 👌 :100:

app/views/examples/example_radios_checkboxes.html
selfthinker: I have removed it because I didn't understand why it was there in the first place. I think it's generally bad practice to have `<br>`s in a heading. I tried to find out via `git blame` why it was there but got lost trying to get before October 2014.

BTW, without this change, it currently looks broken on mobile because of that `br`:
br-in-h1

selfthinker: Ah, I just found it. It was there from the beginning of the checkbox examples in bc2c6b409c10f8f3ab210a1bb5d70ae721b991dd, so there is no further explanation as it was part of a bigger thing. Maybe @gemmaleigh knows more?

LJWatson:
This solution tests well with Jaws in Chrome, Firefox and IE, and with NVDA in Chrome, Firefox, IE and Edge (although the legend itself isn't recognised by NVDA in Edge regardless of whether its in a heading or not).

jbuller:
Let me challenge: Why does the H1 have to match the question?
Usually the H1 should (closely) match the page title attribute (not including the service and GOV.UK names)? 
The title and H1 can be for example 'Your nationality' and the question can be more detailed.
That keeps the non-question style of the titles and H1s the same across the service(s).
I just worry a bit about sowing potential for confusion in to the patterns with exceptions to  consistency for both design teams and users.

LJWatson:
If the title and h1 reflect each other, I think that's ok. So if you have ```<title>Your nationality...</title>```, and ```<h1><legend>What is your nationality</legend></h1>```, I think that's ok.

joelanman:
@jbuller the H1 does not always have to match the question, that's not being proposed. However in many cases it does match, due to the 'one thing per page' pattern. If you look at [Register to vote](https://www.gov.uk/register-to-vote) you can see this in action.

stevenaproctor:
What happens if there is no fieldset and legend but a visually hidden label and a visible form field? Would the h1 go inside the label?

For example, 

<h1>What is your postcode?</h1>
<label class="visuallyhidden">What is your postcode?</label>
<input>

<label><h1>What is your postcode?</h1></label>
<input>...

Is this correct?

Also, for page titles and h1s, HMRC's guidance is they should match. See https://github.com/hmrc/design-language-documentation/issues/90

selfthinker:
If I understand @jbuller correctly, I think he questions the pattern itself to have only the question visible not this code interpretation of it?
For example, if you have more than one thing per page, like your address, the h1 (and half of the title) would be "Your address" and all the inputs would have their own label/legend. If you wanted to keep it consistent, you could keep both h1 and legend/label visible for one thing per page, e.g. "Your account" as the h1 (and half of the title) and "Do you already have an account?" as the legend.
While I think that would make the pattern easier for developers and more consistent, I believe the current pattern (which would have both the h1 inside the legend and half the title as "Do you already have an account?") has been tested multiple times vs the other one and having duplicate information on the page  proved less successful. But designers like @joelanman will know much more about previous user tests and can explain it better.

@stevenaproctor, that's nearly correct, see the code I linked above to [Label in heading](http://jsbin.com/xecoli). So with your example it would be
```
<h1><label>What is your postcode?</label></h1>
<input>...
```
The h1 inside a label would be invalid HTML, so it needs to be the other way around.

stevenaproctor:
Thanks @selfthinker

PR #511, opened 2017-06-29
Change error summary ARIA role to "alert"

selfthinker:

Currently neither VoiceOver on iOS or macOS nor TalkBack on Android announce that there is an error when submitting a form with errors. Although we focus on the error summary box, it doesn't have the desired effect. In the iOS case it's because of [one browser bug](https://bugs.webkit.org/show_bug.cgi?id=163719) and in the macOS case because of [another browser bug](https://bugs.webkit.org/show_bug.cgi?id=171006). At least the iOS and Android issue can be fixed by adding the ARIA `role="alert"` to the error summary box.

I am not sure what the effect will be of removing the `group` role. It didn't seem to have an effect when I tested this but I would like to have someone else's opinion on this.

## How has this been tested?

Changing this doesn't change the behaviour (postively or negatively) in JAWS, NVDA or VO on macOS, but it does mean that ZoomText is not announcing anything.
As VO on iOS is used much more often by blind people than ZoomText (ZoomText is used more by visually impaired people who would be able to see the big red alert box), this seems to be a sensible sacrifice.

## What type of change is it?

### DIFF
--- a/app/views/examples/example_form_validation_multiple_questions.html
+++ b/app/views/examples/example_form_validation_multiple_questions.html
@@ -30,9 +30,9 @@
           </span>
         </li>
         <li>
-          indicate to screenreaders that the summary represents a collection of information
+          ensure that the summary is announced to as many screen readers as possible
           <span class="panel panel-border-narrow">
-            add the ARIA <code>role="group"</code> to the containing <code>div</code>
+            use <code>role="alert"</code> on the containing <code>div</code>
           </span>
         </li>
         <li>
--- a/app/views/examples/example_form_validation_single_question_radio.html
+++ b/app/views/examples/example_form_validation_single_question_radio.html
@@ -36,9 +36,9 @@
           </span>
         </li>
         <li>
-          indicate to screenreaders that the summary represents a collection of information
+          ensure that the summary is announced to as many screen readers as possible
           <span class="panel panel-border-narrow">
-            add the ARIA <code>role="group"</code> to the containing <code>div</code>
+            use <code>role="alert"</code> on the containing <code>div</code>
           </span>
         </li>
         <li>
--- a/app/views/snippets/form_error_multiple.html
+++ b/app/views/snippets/form_error_multiple.html
@@ -1,5 +1,5 @@
 {% if error %}
-<div class="error-summary" role="group" aria-labelledby="error-summary-heading" tabindex="-1">
+<div class="error-summary" role="alert" aria-labelledby="error-summary-heading" tabindex="-1">
 
   <h2 class="heading-medium error-summary-heading" id="error-summary-heading">
     Message to alert the user to a problem goes here
--- a/app/views/snippets/form_error_multiple_show_errors.html
+++ b/app/views/snippets/form_error_multiple_show_errors.html
@@ -1,7 +1,7 @@
 <div class="grid-row">
   <div class="column-two-thirds">
 
-    <div class="error-summary" role="group" aria-labelledby="error-summary-heading-example-2" tabindex="-1">
+    <div class="error-summary" role="alert" aria-labelledby="error-summary-heading-example-2" tabindex="-1">
 
       <h2 class="heading-medium error-summary-heading" id="error-summary-heading-example-2">
         Message to alert the user to a problem goes here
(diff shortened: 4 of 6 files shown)


app/views/examples/example_form_validation_multiple_questions.html
```diff
@@ -30,9 +30,9 @@ <h2 class="heading-medium implementation-advice">Implementation advice</h2>
           </span>
         </li>
         <li>
-          indicate to screenreaders that the summary represents a collection of information
+          make sure that the summary gets announced to as many screen readers as possible
```
36degrees: Minor copy suggestion to improve flow and be consistent with the other points:

> Ensure that the summary is announced to as many screen readers as possible
> use <code>role="alert"</code> on the containing <code>div</code>

36degrees:
Makes sense to me. I'm not an expert on roles by any means, but from a quick Google I don't _think_ we're losing anything important by removing the `group` role.

hannalaakso:
I'm guessing the role role "group" was meant to group related elements (h1, p, ul) together, just looking at form_error_radio_show_errors.html. But [W3 say role "group" should be used for form controls](https://www.w3.org/WAI/GL/wiki/Using_grouping_roles_to_identify_related_form_controls) so I think the group role here is not doing anything, unless there are form controls inside any of the divs which I couldn't see.

selfthinker:
I wonder if the `aria-labelledby` has any symbiotic relationship with the `role=group` and if it can be accessed the same way without it?

@36degrees, the copy change makes sense, I will change it when I am back.

gemmaleigh:
I had a look for the reason why `role="group"` was added, I'm afraid can't find the history of it. I can only assume it was added so that assistive technologies know that it's a container for a group of related content. https://www.w3.org/TR/wai-aria/roles#group

The `aria-labelledby` attribute is there to ensure the heading in the summary is associated with the summary box [when using role="group"](https://www.w3.org/TR/WCAG20-TECHS/ARIA17.html). So when focus is set to the error summary, the heading is announced.

I'd be keen to know if the [JavaScript which is setting focus to the error summary box](https://github.com/alphagov/govuk_elements/blob/master/assets/javascripts/application.js#L25) can also be removed?

The documentation for `role="alert"` implies there isn't a need to also set focus to the summary box: https://www.w3.org/TR/wai-aria/roles#alert
> Neither authors nor user agents are required to set or manage focus to them in order for them to be processed.

hannalaakso:
I've finally got a working machine so can do some testing on this. I'll pick this up tomorrow.

selfthinker:
@gemmaleigh, I have tested lots of variations, but I have not tested what happens if we don't set the focus. Good question. I would guess it's still needed due to inconsistent screen reader support but I can make sure and test it.

@hannalaakso, I can introduce you to our AT laptops and show you how to test in our various screen readers etc.

hannalaakso:
[This](https://www.paciellogroup.com/blog/2017/04/aria-alert-support/) claims that NVDA + Edge doesn't support role="alert". I guess we need to do some testing.

I did test removing the JS that sets focus and using role="alert" and VoiceOver announces it correctly. I was testing in form_error_multiple.html.

Thanks @selfthinker have slacked you separately re:intro 👍

selfthinker:
I don't think NVDA officially supports Edge yet. (Struggling to find evidence of that, though.) So, I think we can ignore that.

selfthinker:
I can currently not fully test this without the focus as our laptops with JAWS etc on them have connection issues.

I have just tested this in NVDA and Firefox, though, and that does definitely **not** work properly without the focus. It jumps straight to the first input field (after reading the page title) and skips the error summary box completely.
Another reason why the focus is still useful is mobile devices. It helps move the box into the viewport!

selfthinker:
I have asked Leonie for feedback on this and she responded:

> We used the group role to set an explicit role on the div element (without which the aria-labelledby attribute won't work). More on this here:
https://www.paciellogroup.com/blog/2017/07/short-note-on-aria-label-aria-labelledby-and-aria-describedby/
>
> If the alert role is used instead it will achieve the same thing, so no problem with switching it over.

I haven't made the copy change @36degrees requested yet because I'm waiting for #509 to be merged as I know there will be conflicts.

selfthinker:
I have just made the copy changes. Although #509 isn't merged yet, this one could be merged before it. I guess it doesn't really matter which one is the one that is causing the other to conflict.

PR #514, opened 2017-07-05
Check for changes to Yarn lockfile

gemmaleigh:

This pull requests adds a script - `check-lockfile.sh` to see if the Yarn lockfile needs updating.
Travis will run this script before running `yarn test`.

This has been tested using Travis in the builds below:

If the lockfile needs updating, the build stage "Run tests" will fail and will output:
"lockfile needs updating, run yarn and commit"
https://travis-ci.org/alphagov/govuk_elements/jobs/250302619#L238

Once the lockfile has been updated, the build runs successfully here:
https://travis-ci.org/alphagov/govuk_elements/jobs/250304438#L237

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,9 @@ jobs:
   include:
     - stage: run tests
       script:
-        # Ensures `package.json` and `yarn.lock` stay in sync
-        - yarn check --integrity
+        # Ensure Travis aborts without running tests if the lockfile needs updating
+        - set -e
+        - ./check-lockfile.sh
         - yarn test
     - stage: heroku deploy
       script: echo "Deploying to Heroku ..."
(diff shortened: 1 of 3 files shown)


check-lockfile.sh
```diff
@@ -0,0 +1,8 @@
+echo "checking Yarn lockfile for changes"
```
36degrees: Totally should've thought about this earlier but I think some context would be useful here. Something like…

> Check to see whether the yarn lockfile was changed when Travis automatically ran yarn. We can't use yarn check --integrity because the lock file will already have been updated (but not committed), and so the check will always pass.

?

.travis.yml
```diff
@@ -9,8 +9,8 @@ jobs:
   include:
     - stage: run tests
       script:
-        # Ensures `package.json` and `yarn.lock` stay in sync
-        - yarn check --integrity
+        - set -e
```
36degrees: Could do with a comment. 'Ensure Travis aborts without running tests if the lockfile needs updating'?

36degrees:
👏

PR #516, opened 2017-07-07
Update govuk-frontent-toolkit name to match npm package naming

abrownhill:
When using the guidance for 'packages/govuk-elements-sass', the README.md file (changed in this pull request) directs to use 'npm install govuk-frontend-toolkit' command however the correct name for the npm package is 'govuk_frontend_toolkit' (just changing hyphens for underscores).

Single change to README.md file therefore no tests executed.

- I have read the **CONTRIBUTING** document. - Apologies though, wasn't sure if this actually needed a new version given that it's a minor, non-functional change.


36degrees:
Thanks for this! In theory that package is installed as a dependency of govuk-elements-sass anyway so not sure how important that line currently is anyway, but making it correct is a step in the right direction!

PR #532, opened 2017-07-14
Make old JAWS recognise disabled buttons

selfthinker:

JAWS 15 and below don't recognise the `disabled` attribute on buttons.
This adds the `aria-disabled` attribute to make disabled buttons compatible with older screen readers.

This does not apply to other `disabled` elements. I also tested the new disabled radio buttons and checkboxes, and they are fine in older JAWS versions without this change.

## How has this been tested?

I tested with JAWS 14, 15 and 17.
JAWS 17 is fine without the change and reads e.g. "Save button - unavailable" but all the others read "Save button" as if it was a normal button, not making it clear that you cannot use it. Adding `aria-disabled` fixes that. It seems redundant but is similar to our practice to add `role="main"` to our `<main>` element to support older browsers.

I couldn't test with JAWS 16 yet. In case that also doesn't support `disabled`, I will update the commit message and this PR accordingly.

## What type of change is it?

<del>I'm not sure if this needs to be documented?</del>
I have added an item to the page.

### DIFF
--- a/app/views/guide_buttons.html
+++ b/app/views/guide_buttons.html
@@ -85,12 +85,13 @@
   <ul class="list list-bullet text">
     <li>don’t disable buttons, unless there’s a good reason to</li>
     <li>if you have to disable buttons, make sure they look like you can’t click them (use 50% opacity)</li>
+    <li>use the <code class="code">aria-disabled</code> attribute for older screen readers</li>
     <li>an example of a useful disabled option is a calendar with greyed out days where no bookings are available</li>
     <li>provide another way for the user to continue, add an error message or text to explain why the button is disabled</li>
   </ul>
 
 <div class="example">
-  <button class="button" disabled="disabled">Disabled primary button</button>
+  <button class="button" disabled="disabled" aria-disabled="true">Disabled primary button</button>
 </div>
 
 <pre>
--- a/app/views/snippets/buttons_button_disabled.html
+++ b/app/views/snippets/buttons_button_disabled.html
@@ -1,1 +1,1 @@
-<input class="button" type="submit" value="Save and continue" disabled="disabled">
+<input class="button" type="submit" value="Save and continue" disabled="disabled" aria-disabled="true">


gemmaleigh:
Thanks @selfthinker!

For the documentation - could you please add another list item to the list of bullet points about disabled buttons:

* disabled buttons must use the`aria-disabled` attribute for older screen readers

selfthinker:
I have added an item to the list. I have shortened your proposed sentence because we rarely use the word "must" (or "should" etc) within Elements and I thought it sounded a bit too formal for that bit.

selfthinker:
Unfortunately, we can currently not test in JAWS 16. I've asked in the x-gov slack channel if anyone can test it. Otherwise, I don't think it's super important. It would just be good to know which JAWS version started to support the `disabled` attribute on buttons. But this could be merged without that knowledge. Let's wait until tomorrow.

selfthinker:
@accessiblewebuk just tested and confirmed that JAWS 16 handles the buttons correctly, so it is indeed "JAWS 15 and below" as stated in the commit message and description above.

edwardhorsford:
Would it be good to be specific about what screen readers and versions we're doing this for? I feel like this could be helpful in some years time when we can take a view if we still support those screen readers.

selfthinker:
I was specific in the commit message ("JAWS 15 and below"), which people should always be able to find. Or did you mean making it more prominent in the guidance?

edwardhorsford:
I mean making it explicit in the guidance - but it's only a suggestion.

gemmaleigh:
I'm happy to merge, before I do - let me know if you would prefer to use "older screen readers" or "JAWS 15 and below".

selfthinker:
We don’t necessarily know that it’s just older JAWS as we haven’t tested older versions of other screen readers. (Although the reason for that is that older versions of other screen readers are not used as much.)
We are also not specific in other areas either (e.g. with `role="main"` or many JavaScripts), so that would be inconsistent. Although, it would still be a good idea to generally do it, but it might be quite a lot of work to do it consistently.

gemmaleigh:
Thanks @selfthinker, will merge as-is.

PR #534, opened 2017-07-21
Remove tabindex from main element

fofr:
Reverts https://github.com/alphagov/govuk_elements/pull/225
See: https://github.com/alphagov/govuk_template/pull/321

Putting tabindex on the `<main>` element causes different problems:
* Some apps will display the browser's default focus styles around the main element
* When clicking anywhere in the page focus will return back to the top

Consider a user interacting with an input field who clicks away from it to remove
the focus style. Should they then hit tab they will be taken to the top of the page.

The [linked to browser bug](https://bugs.chromium.org/p/chromium/issues/detail?id=37721) was fixed in Apr. Testing on Safari 10.1.1 it works with VoiceOver on desktop. Testing on iOS 10.2.1 it works in mobile Safari with VoiceOver on.

More context around the issue:
https://github.com/twbs/bootstrap/issues/20732

## Example problem
GIF showing focus styles and tab order problem on Registers:

focus-order-issue-tabindex-main

### DIFF
--- a/app/views/examples/example_date.html
+++ b/app/views/examples/example_date.html
@@ -3,7 +3,7 @@
 {% block page_title %}Example: Date — Form elements — GOV.UK elements{% endblock %}
 
 {% block content %}
-<main id="content" role="main" tabindex="-1">
+<main id="content" role="main">
 
   {% include "includes/breadcrumb.html" %}
 
--- a/app/views/examples/example_details_summary.html
+++ b/app/views/examples/example_details_summary.html
@@ -3,7 +3,7 @@
 {% block page_title %}Example: Details summary — Typography — GOV.UK elements{% endblock %}
 
 {% block content %}
-<main id="content" role="main" tabindex="-1">
+<main id="content" role="main">
 
   {% include "includes/breadcrumb.html" %}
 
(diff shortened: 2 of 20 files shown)


selfthinker:
It looks like all the browser vendors have fixed the bug that was the reason for this. But it would be good to have a list of all the browsers this was tested in.

gemmaleigh:
I commented over on GOV.UK template - this was added as a fix to ensure the target of the skiplink was focussed. If this is no longer an issue - I'm happy for this code to be deleted.

fofr:
@gemmaleigh As we don't control the HTML of pages, should we leave in the outline rule within layout.scss?

gemmaleigh:
@fofr can we leave it in for now and add a comment that the tabindex attribute is no longer required, then we can remove it in a later release?

nickcolley:
@gemmaleigh 👍  @fofr can you update please? This seems fairly important would be good to get a new release out.

fofr:
@gemmaleigh @nickcolley `outline` kept in but deprecated in 5bdc062

gemmaleigh:
thanks @fofr 👍

selfthinker:
And no-one has checked in which browsers it actually works? I would think that's vital.

fofr:
@selfthinker IMO the bug this introduced was worse than the one it fixed. Which makes it _less_ vital.

Tested on:
* Safari 10.1.1 it works with VoiceOver on MacOS Sierra
* iOS 10.2.1 and 10.3.3 it works in mobile Safari with VoiceOver on

These were the browsers the bug fix appeared to target.

PR #547, opened 2017-08-15
Adjust the icon position for legal text

gemmaleigh:
### What problem does the pull request solve?

Ensure that the icon sits aligned with the middle of the first line of
text, rather than vertically centred for the entire block of text.

Adjust the position for larger viewports, when the text size increases
from 16px to 19px.

#### Before:

legal text - before - typography gov uk elements

#### After: 

legal text - after - typography gov uk elements

---

For smaller viewports when the text wraps:

small screen - typography gov uk elements

---

For larger viewports with a single line of text:

wide screen - typography gov uk elements

### DIFF
--- a/assets/sass/elements/_elements-typography.scss
+++ b/assets/sass/elements/_elements-typography.scss
@@ -286,8 +286,12 @@ hr {
   .icon {
     position: absolute;
     left: 0;
-    top: 50%;
-    margin-top: -17px; // Half the height of the important icon
+    top: 0;
+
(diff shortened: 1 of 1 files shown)


trevorsaint:
Brilliant Gemma ;)

edwardhorsford:
Coming late to this party. Any reason it's not top-aligned to the top line of text?

I also wonder if it should scale with the font-size - but guess that would be a separate PR.

igloosi:
@edwardhorsford  yeah, the latter would be an svg solution

Issue #553, opened 2017-09-14
Replace fragile Firefox workaround for <summary> arrow icon


robinwhittleton:
The styling for `<summary>` on GOV.UK currently expects the focus ring to shrinkwrap the text. This works fine in Chrome, but fails in Firefox. I worked around this with some [horrendous code](https://github.com/alphagov/govuk_elements/blob/dea6e7bbaf83ecc670bc2cd31539096fab542484/assets/sass/elements/_shame.scss) but:

1. that just resets Firefox back to not shrinkwrapping, and
2. it turned out that the Firefox implementation is based on [the WHATWG suggested initial rendering](https://html.spec.whatwg.org/multipage/rendering.html#the-details-and-summary-elements) for summary.

Having read through I started worrying that other browsers might implement the WHATWG initial styles in the future, but a different way of fixing the problem did occur to me. While we can’t use `display: list-item` directly, we can include the list icons directly using counters. The following example code seems to work in Firefox and Chrome at least, although it’s untested beyond that and would need a proper accessibility run through etc:

```css
summary {
  display: inline-block;
  counter-reset: summary-closed;
}
summary::before {
  content: counter(summary-closed, disclosure-closed) " ";
}
details[open] > summary {
  counter-reset: summary-open;
}
details[open] > summary::before {
  content: counter(summary-open, disclosure-open) " ";
}
```

`disclosure-closed` and `disclosure-open` are the default markers used for `<summary>` elements. The `" "` just inserts an empty space inbetween the marker and the text.

Does this sound like a reasonable approach?

36degrees:
This was done in https://github.com/alphagov/govuk_elements/pull/611.

PR #554, opened 2017-09-15
Fix summary outline shrink-wrapping in Firefox

robinwhittleton:
The styling for `<summary>` on GOV.UK currently expects the focus ring to shrinkwrap the text. This works fine in Chrome and Safari, but fails in Firefox (see https://github.com/alphagov/govuk_elements/issues/553).

This removes the old workaround to get the arrows to show in Firefox. It then reimplements the arrows for everyone using counters. There is a potential problem that browsers that understand `counter` but not `<details>` might get another arrow in addition to the one generated by the polyfill. As luck would have it though, the list styles `disclosure-open` and `disclosure-closed` seem to have been invented for the details element, so the versions of Edge and IE that I tried this fix in don’t recognise the `::before`/`::after` `content` rules and don’t show them.

Tested in latest Firefox / Chrome / Safari / Edge and IE11. Has not been tested on mobile, other browsers or with assistive tech.


igloosi:
Screenshots from Firefox: 
Before:
image

After:
image

Thoughts from the design community?

igloosi:
There's 3 ways we can approach this
1) we have this slightly smaller arrow in Firefox (we can adjust padding) and have Safari and chrome use details-marker (plus polyfill for other browsers, that use \u25bc and \u25ba)

2) forcefully hide details-marker for Safari and Chrome and make them use \25B6 so that all browsers, including polyfill will the same icon. Behaviour of it is strange, it pushes text right, see here https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details

3) replace <details> with button + javascript (we already have js polyfill for the element)
Question here is, is details the best way to display disclosure content? Does it have better browser accessibility built in than button + content combo? I would assume so
http://accessibleculture.org/articles/2012/03/screen-readers-and-details-summary/

Input here welcome
@selfthinker @edwardhorsford

edwardhorsford:
I don't feel able to comment too much on the options other than that we wouldn't want the text to move when opening and closing the item.

igloosi:
Hi all,
in Frontend we've taken a slightly different approach
https://github.com/alphagov/govuk-frontend/pull/372

robinwhittleton:
I’m happy for this to be closed then.

36degrees:
Superseded by #611, which mirrors the approach we've taken with GOV.UK Frontend.

PR #556, opened 2017-09-25
Document invalid HTML in date pattern.

### LINKED ISSUE #551: Date form example uses invalid HTML
The [date form example](https://govuk-elements.herokuapp.com/form-elements/example-date/) uses invalid HTML. The [W3C Validator](https://validator.w3.org/nu/?doc=https%3A%2F%2Fgovuk-elements.herokuapp.com%2Fform-elements%2Fexample-date%2F) says:

> Attribute `pattern` is only allowed when the input type is `email`, `password`, `search`, `tel`, `text`, or `url`.

It was added in #47, and I believe it was done to make sure the number keyboard shows up in some devices -- at least at the time.

We should find out if it's still necessary, and if it is, we should add some kind of documentation somewhere that the invalid HTML is intentional.

hannalaakso:
This documents the reason for invalid HTML in the date pattern:
A HTML validator will produce the following error for this date pattern: "Attribute pattern is only allowed when the input type is email, password, search, tel, text, or url." This is a known issue. The pattern attribute is needed to force numeric keypads on iOS devices. Type=number alone does not force the numeric keypad. Take a look at [this article](http://bradfrost.com/blog/mobile/better-numerical-inputs-for-mobile-forms/) which explains the issue in more detail.

@selfthinker: do you know if this has been reported to W3?  

Fixes #551 

Tested in Chrome/FF/Safari.

- Added an explanation to a view using existing elements.

### DIFF
--- a/app/views/examples/example_date.html
+++ b/app/views/examples/example_date.html
@@ -97,13 +97,8 @@
     </div>
     <div class="panel panel-border-wide text">
       <p>
-        A HTML validator will produce the following error for this date pattern: "Attribute pattern is only allowed when the input type is email, password, search, tel, text, or url."
-      </p>
-      <p>
-        This is a known issue. The pattern attribute is needed to force numeric keypads on iOS devices. Type=number alone does not force the numeric keypad.
-      </p>
-      <p>
-        <a href="http://bradfrost.com/blog/mobile/better-numerical-inputs-for-mobile-forms/" target="_blank">Take a look at this article</a> which explains the issue in more detail.
+        Note: validators will identify the <span class="bold">pattern</span> attribute as not being valid for inputs
+        where the <span class="bold">type</span> attribute is <i>number</i>. It is added to <a href="http://bradfrost.com/blog/mobile/better-numerical-inputs-for-mobile-forms" target="_blank">force numeric keypads on iOS devices</a>.
       </p>
     </div>
   </div>
--- a/app/views/examples/example_date.html
+++ b/app/views/examples/example_date.html
@@ -97,8 +97,7 @@
     </div>
     <div class="panel panel-border-wide text">
       <p>
-        Note: validators will identify the <span class="bold">pattern</span> attribute as not being valid for inputs
-        where the <span class="bold">type</span> attribute is <i>number</i>. It is added to <a href="http://bradfrost.com/blog/mobile/better-numerical-inputs-for-mobile-forms" target="_blank">force numeric keypads on iOS devices</a>.
+        Note: The <code class="code">pattern</code> attribute is not valid HTML for inputs where the <code class="code">type</code> attribute is <code class="code">number</code>. It is added deliberately to <a href="http://bradfrost.com/blog/mobile/better-numerical-inputs-for-mobile-forms">force numeric keypads on iOS devices</a>.
       </p>
     </div>
   </div>
--- a/app/views/examples/example_date.html
+++ b/app/views/examples/example_date.html
@@ -95,6 +95,17 @@
       </div>
 
     </div>
+    <div class="panel panel-border-wide text">
+      <p>
+        A HTML validator will produce the following error for this date pattern: "Attribute pattern is only allowed when the input type is email, password, search, tel, text, or url."
+      </p>
+      <p>
+        This is a known issue. The pattern attribute is needed to force numeric keypads on iOS devices. Type=number alone does not force the numeric keypad.
+      </p>
+      <p>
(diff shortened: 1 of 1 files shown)


edwardhorsford:
Thanks for adding this @hannalaakso. I wonder if we could say it in a more concise way? maybe have Amy / Ellen look at it?

hannalaakso:
Thanks Ed, I'll ask them to review it.

selfthinker:
I've tried re-writing it. What do you think about this?

> Please note, this pattern intentionally includes invalid HTML. The `pattern` attribute is not allowed on `number` input, but it is needed to [force numeric keypads on iOS devices](http://bradfrost.com/blog/mobile/better-numerical-inputs-for-mobile-forms/).

Also, I think using the panel style doesn't fit here. Or at least I'm not aware we're using it in any similar situation. The problem with the page is that it is quite long and the text might be overlooked. Maybe it's worth putting it at the top above the first example (without the panel style, just a normal paragraph)?

selfthinker:
Regarding the question if this has been reported to the W3C, I don't think there is anything wrong with it. Which kind of patterns would you want to have on a number for which you cannot use `max`, `min` and/or `step` and which are true numbers?
In our case you wouldn't need the `pattern` if Safari wasn't buggy.

hannalaakso:
I've rewritten the content with Ellen and @36degrees. This style/format could hopefully be used across elements for documenting similar issues.

@selfthinker Re: the panel style, I'm not too offended by it by I know what you mean. There doesn't seem to be any other elements available for this. @edwardhorsford do you think we should take this to a designer to see if we require a new style for this type of notification?

selfthinker:
I don't like that it's written like it's blaming validators. If validators tell us about the invalid HTML or not shouldn't be important. What's important is that it's invalid HTML.

As for the panel style, I guess it's best a designer looks at that.

app/views/examples/example_date.html
```diff
@@ -97,13 +97,8 @@ <h1 class="heading-xlarge">
...
       <p>
-        A HTML validator will produce the following error for this date pattern: "Attribute pattern is only allowed when the input type is email, password, search, tel, text, or url."
-      </p>
-      <p>
-        This is a known issue. The pattern attribute is needed to force numeric keypads on iOS devices. Type=number alone does not force the numeric keypad.
-      </p>
-      <p>
-        <a href="http://bradfrost.com/blog/mobile/better-numerical-inputs-for-mobile-forms/" target="_blank">Take a look at this article</a> which explains the issue in more detail.
+        Note: validators will identify the <span class="bold">pattern</span> attribute as not being valid for inputs
+        where the <span class="bold">type</span> attribute is <i>number</i>. It is added to <a href="http://bradfrost.com/blog/mobile/better-numerical-inputs-for-mobile-forms" target="_blank">force numeric keypads on iOS devices</a>.
```
selfthinker: Why do you use `<span class="bold">` and `<i>`? We usually use `<code class="code">` for these things.

selfthinker: We usually never use `target="_blank"` (except in rare cases) as it is not very user-friendly. I would have added a link to the guidance, but I cannot find it right now. @edwardhorsford, do we have something official anywhere?

robinwhittleton: There’s info on https://paper.dropbox.com/doc/Links-xiWNcJk5Wpb8Sj6G4Prtl

hannalaakso:
Thanks @selfthinker. I have amended the code tags, removed target="_blank" and amended the note content together with a content designer.

@dashouse: How do you feel us about using the panel element for the note about code [here](https://govuk-elements-date.herokuapp.com/form-elements/example-date/)? Is there another element better suited for this?

dashouse:
Hi, I think that's a valid use of inset text, however it should be inside the `column-two-thirds` div as it's currently sitting outside of the main layout.

36degrees:
@hannalaakso it'd be good to remove that merge from the master branch to keep the history simpler. Let me know if I can help with that.

hannalaakso:
Thanks @36degrees that's now done. 

Thanks @dashouse I've corrected the column issue, see [here](https://govuk-elements-date.herokuapp.com/form-elements/example-date/)

selfthinker:
I don't think that is what @dashouse meant with the `column-two-thirds` div. It should have been inside the div that was already there, not in a new one.

In the future, I would suggest to merge commits into logical steps. In this case I would have just created 1 commit and not 4 as they are all about the same change and the subsequent commits are about changing the existing change and not adding anything new. Keeping commits about one logical change will make it easier and quicker for people to read and follow the git history.

PR #560, opened 2017-09-29
Should our colour naming be unopinionated

charlesrt:
Some of our colour names are opinionated.

This pull request changes them to be neutral, and also correctly spells fuchsia.

screen shot 2017-09-29 at 17 35 50

This change could be breaking if any of these Sass variables are being used.


app/views/guide_colour.html
```diff
@@ -234,12 +234,12 @@ <h4 class="heading-small">Mauve</h4>
         <li>$mauve</li>
       </ul>
 
-      <h4 class="heading-small">Fuschia</h4>
```
fofr: Good catch. Might be worth splitting this change into a separate PR so it can be merged relatively quickly without some of the expected debate associated with the other changes.

fofr: (Or we pick an easier to spell colour name – not that I can think of an alternative)

36degrees:
I don't know if this warrants a breaking release or if this is something we can address as part of the rollout of GOV.UK Frontend instead?

If we do want to change this in Elements there will need to be a corresponding PR for Frontend Toolkit to change the [variable definitions themselves](https://github.com/alphagov/govuk_frontend_toolkit/search?utf8=%E2%9C%93&q=fuschia&type=) and we'll need to release a breaking version of that as well, unless we define both but mark the former as deprecated?

The tests are currently failing because of this – `$fuchsia` is not actually being defined.

nickcolley:
As this PR stands this would definitely be a breaking release, so would need to be backwards compat with deprecation warnings.

charlesrt:
@fofr That's a good idea to make the colour naming simpler too. 

@nickcolley You're right, this doesn't need to be a breaking change if it the current Sass variables are stay supported.

@36degrees I've confused myself as to how it all hangs together. The changes needs to happen in the frontend toolkit in _palette.scss which is pulled in. The only change to elements would be content.

I'll close this PR.

joelanman:
Suggest 'plum' - easier to spell

charlesrt:
@joelanman that's a good idea.

I raised this as an issue #561 to make debate possible (I probably should have done that first).

PR #568, opened 2017-10-17
Replace 'lorem ipsum' with English

selfthinker:
We currently use filler text in some instances on the typography page. This replaces that with English. It also improves the copy of the various list items.

Because some of the text was duplicated in two files each, I also added a filter to use just one of the files instead of two.
Unfortunately a filter block would not allow to chain multiple filters (e.g. `{% filter replace('foo', '') | replace('bar', '') %}`), so I needed to do it in a bit of a roundabout way.

### DIFF
--- a/app/views/guide_typography.html
+++ b/app/views/guide_typography.html
@@ -56,7 +56,7 @@
   </ul>
 
 <div class="example">
-  {% include "snippets/typography_headings_example.html" %}
+  {% include "snippets/typography_headings.html" %}
 </div>
 
 <pre>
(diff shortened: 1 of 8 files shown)


36degrees:
Playing devil's advocate a little here, but I'm wondering how useful the highlights actually are – especially as they look a lot like links?

screen shot 2017-10-25 at 09 48 22

selfthinker:
@36degrees, I think that's a fair point, I wondered the same.
I just went through the git history to find out if there was an explanation why they are there, but the highlighting was added together with the snippets the first time they were added (in #22).
Maybe anyone on that original PR can remember? (Only @joelanman is still here.)

I'm happy to change the PR to remove them. I think it makes sense. I'll wait 24 hours with that in case someone disagrees.

selfthinker:
@36degrees, I just removed the highlighting altogether.

Issue #357, opened 2016-11-21
Add hover states back in for radios / checkboxes after iOS < 9.1 goes away


robinwhittleton:
The new radios / checkboxes had [hover states](https://github.com/alphagov/govuk_elements/pull/296/commits/a844cd2a472b82e84306f7a2c4fc5a0990c01cbb#diff-05868dd1ba9cecec849b639448452f51R93) during their development and @edwardhorsford asked why it was removed. It had problems in iOS due to it’s ‘sticky’ hover behaviour showing ghost checkmarks after deselection. We managed to solve it using CSS4 hover media queries: https://github.com/alphagov/govuk_elements/pull/296/commits/e654d8bc721b0eba11bd11010600fa33da1dd077 . Unfortunately this was only Safari 9.1+, and we decided that the bad outweighed the good and removed it.

We should keep an eye on old iOS and add the hover behaviour back in again when Safari <9.1 mostly goes away.

robinwhittleton:
The GOV.UK browsers and devices page currently recommends iOS Safari 9.3+; is this worth re-evaluating given the usability improvements for other browsers? I don’t have access to the stats to make the call, so would need to be discussed inside GDS.

joelanman:
thanks robin! That sounds like the elements will fade out (opacity 0.15) when hovered - is that right? You got a screenshot anywhere?

robinwhittleton:
Yep, it looked like this:
<img width="418" alt="screen shot 2017-10-27 at 10 59 56" src="https://user-images.githubusercontent.com/7414/32096235-22189ada-bb06-11e7-8376-74f3e76cf63e.png">
Checkboxes were similar; a 0.15 opacity checkbox while hovering.

joelanman:
thanks! So thats when it isn't selected yet. Is there a change to a selected radio when hovered?

robinwhittleton:
Not to the radio, no, as you can‘t deselect a radio. You can deselect a checkbox, but I’m not sure it would be useful to have a hover state there; the main point of the hover state is to give a hint of meaning to what could be just an empty box.

edwardhorsford:
We could try it, but I'd be worried that a hover state on selected items might make them look unselected again. Eg: you pick something, move the mouse a little, and then it looks unselected. If we wanted a hover effect for that, we might want to look at something that doesn't apply it to the check itself but some other element (border or whatever).

joelanman:
[...]

PR #570, opened 2017-10-28
Add a hover state for radios and checkboxes

robinwhittleton:
This gives radio buttons and checkboxes a faint grey hover state, except in the following situations:

1. Where the input is already selected
2. Where the input is disabled
3. Where the user is on a mobile device (Android Chrome, iOS9.1+)

We originally had this as part of the initial custom radios / checkboxes work, but it was removed. The `@media hover` block wasn’t supported in iOS 9.0 and lower, and without that support ghost checkmarks would remain in place after unchecking them. This was decided to be too off-putting to warrant inclusion. Now that GOV.UK’s browsers and devices page suggests that iOS 9.3+ can be considered a reasonable minimum, we can consider adding hover back in again as a usability enhancement.

Fixes https://github.com/alphagov/govuk_elements/issues/357 .

Tested in latest Firefox, Chrome and iOS11 (Simulator). There’s nothing particularly complicated here, and I can’t see any particular problems with support for this. Firefox for Android, IE mobile and iOS9.0 and lower don’t support the `hover` media query, so unchecking a checkbox will leave a ghost checkmark in place until another click happens elsewhere in the page. The usage of these browsers will need to be balanced against the usability enhancement for other browsers. IE9+ should support all the main CSS here without problem.

This is the behaviour with the patch:
An image showing the hover behaviour described above.

### DIFF
--- a/assets/sass/elements/forms/_form-multiple-choice.scss
+++ b/assets/sass/elements/forms/_form-multiple-choice.scss
@@ -111,6 +111,17 @@
     @include box-shadow(0 0 0 3px $focus-colour);
   }
 
+  // Hover state
+  input:not(:checked):not(:disabled):hover + label::after {
+    @include opacity(0.15);
+  }
+  // hide the hover state in iOS9.1+ / mobile Chrome
+  @media (hover: none) {
+    input:not(:checked):not(:disabled):hover + label::after {
+      @include opacity(0);
+    }
+  }
+
   // Selected state
   input:checked + label::after {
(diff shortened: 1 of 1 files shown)


nickcolley:

I spoke to @timpaul about this, to see if this is something we should consider for the GOV.UK Design System.

Since this is an interesting addition to these components but since this does not seem to be backed by research we're going to close this out and consider it for the future if the need arises.

That said this work is great, thanks for taking the time to contribute.

PR #579, opened 2017-11-29
Add the Google Webmaster Tools verification code for better monitoring

Nooshu:
Issue raised this morning where a broken set of pages were on live. These were reported by an external party. Adding in the meta tag in the head to verify the site. We can then setup email alerts using Webmaster Tools to notify us when we have errors on the live site.

Simple meta tag addition to the head element. Change will have no effect on site functionality.

No documentation change required. Comment in code explaining the additions meaning.


joelanman:
@Nooshu do we need to do the same for the prototype kit?

Nooshu:
@joelanman yes good idea. I will set it up for elements, then do the same for the prototype kit too.

PR #580, opened 2017-12-06
[DO NOT MERge] Add txt file to reclaim Google Analytics account

igloosi:
We need to reclaim a google analytics account. To do this a file needs
to be added to the root of the site.

As we can't direct upload files to heroku and our `public/` folder isnt
checked in, so this is a workaround.

Once the account is verified, the file and amended tasks will be
deleted.

Locally with the `heroku-postbuild` task that runs `gulp build` task

[Design System Trello ticket](https://trello.com/c/STpEjt79/459-add-txt-file-to-elements-so-we-can-claim-the-account)

### DIFF
--- a/app/routes.js
+++ b/app/routes.js
@@ -254,5 +254,10 @@ module.exports = {
       var pageName = 'Alpha and beta banners'
       res.render('guide_alpha_beta', { 'page_name': pageName })
     })
+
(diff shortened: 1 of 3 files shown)


igloosi:
true, will do that

igloosi:
account reclaimed. closing

PR #584, opened 2017-12-13
Rename "legal text"

selfthinker:
This is about clarifying what is currently called ["legal text"](https://govuk-elements.herokuapp.com/typography/#typography-legal-text).

This is a follow-up to the discussion on #564 and discussions I had with @timpaul and @stephengill separately.

The name "legal text" is confusing. It is meant to highlight important text and is used as that by most people. The PR which introduced it (#116) had started it as "warning text" and GOV.UK (Govspeak and other GOV.UK Components) are also calling it "warning callout".
It was called "legal text" to avoid being misused and over-used. But it didn't work out as there are cases where it is simply used to highlight something. If it is not used as intended, it makes sense to use a broader term and to clarify the use.

Yes, this change is mostly about the documentation.


edwardhorsford:
I'd like to suggest this be split in to two separate PRs. One for the warning text, and one for the inset text. I'd like further discussion of the inset text (it's used extensively across GOV.UK), but that need not hold up harmonising language on warning text.

selfthinker:
Yes, that's a good idea. I will change this one into the warning text issue and will create a new one for the inset text issue.

selfthinker:
I have split this now up into not just two but three different PRs. The other ones are #586 and #587.

edwardhorsford:
I'd prefer to call it 'warning text'. I agree legal is too limiting, but also feel that 'important' is too broad and vague.

If we go with 'important' is the exclamation mark too extreme?

robinwhittleton:
Consequential text? I.e. there will be consequences to bear in mind.

cjforms:
I assume that for historic reasons, we will end up with at least one of these methods - at least for a while - and this one seems to me to be the one that is likely to survive for longest. 

If the thing comes with an exclamation icon, I'd recommend calling it 'exclamation icon text', or possibly 'exclamation text' for short. 

As with my comment on the #587 pull request, please can the example show what this looks like in the context of a page. And also, please can it have 'if' before 'then' in the example, like this:
"If you don't register then you can be fined up to £5,000"
Or, in line with the results from our user research with designers and developers back in 2014, have an example that makes it clear that the text IS an example, such as:
"If you don't change this example in your service then it will confuse your users"
(maybe that's too frivolous - not meaning to be)

timpaul:
In the interests of delivering, we're going for 'Important' not 'Warning'. It's broader yes, but I think in a useful way.

36degrees:

timpaul:
Ignore my last post, we're going with 'Warning text' now (Ed made a very good case). Thanks everyone for your patience.

selfthinker:
Thanks @timpaul. I have just made the change and updated the 'after' screen shot in the description.

selfthinker:
@timpaul, yes, that's already changed in this pull request. I searched for "legal text" (and similar) and changed every occurrence as well as adding a redirect from the old to the new anchor.

cjforms:
Is there a reason why the stronger part of the instructions ("Use this sparingly") appears as the second point rather than as the first point?
Is there a reason why the 'use this....' paragraph has 'if' part-way along it?
Is there a reason why the 'if' is still part-way down the sentence in the example? 
Please could we have a 'research needed' thing somewhere, as this has led to so much discussion?

Here's my suggestion for the 'after' example:
/----start of suggestion
**Warning text**
Use this sparingly. There is some research evidence that users may not understand this element in the way we intend it.

If you need to warn users about something important, for example legal consequences of an action, then use an exclamation icon followed by bold text.

Also check that the underlying hidden text (in this case 'warning') matches the visible text. 
/--- end of suggestion

selfthinker:
The reason is: I'm not a content designer. ;-)
I'm happy to make changes to the text. @stephengill and @timpaul, what do you think?

I think the research is not needed so much around this particular pattern but much more about the other one (the "inset text").

cjforms:
On the reason: very good point!
Looking forward to  @StephenGill and @timpaul, comments on my quick draft
On the research: may as well research them together, as they are similar patterns in many ways.

selfthinker:
Yes, good point about killing two birds with one stone and testing the patterns together.

PR #587, opened 2017-12-13
Don't highlight text within documentation

selfthinker:
There should not be a need to highlight text within the GOV.UK Elements documentation.
This removes all 6 occurrences of inset text used to highlight content within the documentation. Even before deprecating it, I noticed that none of it really makes any sense to highlight.

This change would be necessary if we merged #586 but also makes sense on its own.

## Bottom note on summarising errors

(I have also changed the wrongly used heading to a paragraph.)

## Form error example page with additional guidance

(This is the only case where I think it could make sense to keep it.)

### DIFF
--- a/app/views/examples/_form_validation_implementation_advice.html
+++ b/app/views/examples/_form_validation_implementation_advice.html
@@ -9,25 +9,16 @@
     show an error summary at the top of the page
   </li>
   <li>
-    move keyboard focus to the start of the summary
-    <span class="panel panel-border-narrow">
-      to move keyboard focus, put <code>tabindex="-1"</code> on the containing div and use <code>obj.focus()</code>
-    </span>
+    to move keyboard focus to the start of the summary, put <code>tabindex="-1"</code> on the containing div and use <code>obj.focus()</code>
   </li>
   <li>
-    ensure that the summary is announced to as many screen readers as possible
-    <span class="panel panel-border-narrow">
-      use <code>role="alert"</code> on the containing <code>div</code>
-    </span>
+    to ensure that the summary is announced to as many screen readers as possible use <code>role="alert"</code> on the containing <code>div</code>
   </li>
   <li>
     use a heading at the top of the summary
   </li>
   <li>
-    associate the heading with the summary box
-    <span class="panel panel-border-narrow">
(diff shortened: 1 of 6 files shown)


edwardhorsford:
Would it be possible to add a before / after screenshot?

selfthinker:
I really don't think making screenshots makes sense as you would need to see everything within context. (And it would result in 12 very long screenshots.) Instead I've just listed all the relevant sections to the description and added before and after links to them.

cjforms:
I agree with this pull request and I think it gives some useful examples of places where inset text doesn't do anything much at all. 

Will the links mentioned in this pull request (especially the 'after' ones) continue to be available indefinitely? If so, then I agree that there's no need for screenshots. If not, then please can we have screenshots, no matter how many of them or how long they are, so that they continue to be available as an example? 

By the way, on this page: https://govuk-elements-review-pr-587.herokuapp.com/errors/#summarise-errors, I spotted an errant 'should' that somehow sneaked past me when we wrote this originally:

"Additionally, you should reflect that there's been an error in the <title> by prefixing the existing title with "Error: ". That will make sure screen readers are alerted to there being an error as soon as possible."

Two options for fixing this:
1. Change 'should' to 'must' (simplest)
OR
2. Change the whole paragraph around to avoid 'must reflect' which sounds a little awkward to me. Suggestion 
"To make sure that screen readers are alerted to the error as soon as possible, you must prefix the existing title with "Error: "
(which has the merit of being shorter, too).

36degrees:

selfthinker:
Hmm, even if the 'after' links still work afterwards, it will be the 'before' that will be missing a reference, because the current version will be updated to be the 'after' version.
I guess that means it definitely makes sense to add screenshots. I will do that later.

selfthinker:
I have just added before and after screenshots to the description.
I also rebased with master, so the checks are not failing anymore.

selfthinker:
@edwardhorsford and @timblair, is there anything in the way of merging this? If yes, what exactly?

nickcolley:
Following the launch of the GOV.UK Design System, GOV.UK Elements will now only get major bug fixes and security patches.

I believe in this case, we deprecated inset text, but since then we have decided to continue recommending the [inset text component](https://design-system.service.gov.uk/components/inset-text/).

Thanks for raising this originally, sorry for the lack of updates.

PR #591, opened 2018-01-10
Fix misaligned radio example

### LINKED ISSUE #581: Radios with headings inside labels are misaligned
Pointed out by gordonmcmullan_hmrc this example from Elements is misaligned

screen shot 2017-12-07 at 16 06 45

I believe this was affected by this PR (https://github.com/alphagov/govuk_elements/blob/8f88ccc46f7e71d2c8f4a56f4bac4beda1126cd5/assets/sass/elements/_elements-typography.scss) to add `display: block` to headings. It looks like that the margins we're previously collapsing, but with `display: block;` they are not.

**Without `display: block;`**
screen shot 2017-12-07 at 16 39 19

Would suggest contextually adjusting the heading classes within the labels rather than removing the `display: block;`

selfthinker:
Since `display: block` was introduced for all heading classes in 8f88ccc46f7e71d2c8f4a56f4bac4beda1126cd5 the [radio example with 'headings' inside labels](https://govuk-elements.herokuapp.com/form-elements/example-radios-checkboxes/) was visually broken.
This fixes #581 by switching the `heading-small` class with the more fitting `form-label-bold` class.

## What type of change is it?

### DIFF
--- a/app/views/examples/example_radios_checkboxes.html
+++ b/app/views/examples/example_radios_checkboxes.html
@@ -130,7 +130,7 @@
             <div class="multiple-choice">
               <input id="radio-part-2" type="radio" name="housing-act" value="Part 2">
               <label for="radio-part-2">
-                <span class="heading-small">Part 2 of the Housing Act 2004</span><br>
+                <span class="form-label-bold">Part 2 of the Housing Act 2004</span>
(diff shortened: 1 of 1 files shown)


cjforms:
Nice change, especially as the class name reflects what's happening

PR #595, opened 2018-02-13
Replace Hackpad links with Dropbox Paper

### LINKED ISSUE #594: Links to DesignPattern Hackpad are broken
There are currently 11 links to the old [DesignPattern Hackpad](https://designpatterns.hackpad.com/) which are now broken because they don't redirect correctly to the new [Dropbox Paper](https://paper.dropbox.com/doc/GOV.UK-Design-Patterns-Wiki-hkou7bpzKya8Dadv24V1z).

I tried to fix it quickly but the content has diverged so much that there are no 1:1 equivalents anymore. (I couldn't find anything about "colour" or conditional form fields, radio buttons and checkboxes are now two different pages, etc.)

igloosi:
Updates broken links.
Fixes #594.

Links checked.

This is a documentation update.

### DIFF
--- a/app/views/includes/phase_banner.html
+++ b/app/views/includes/phase_banner.html
@@ -1,6 +1,6 @@
 <div class="phase-banner">
   <p>
     <strong class="phase-tag">ALPHA</strong>
-    <span>This is a new service – your <a href="https://designpatterns.hackpad.com/GOV.UK-elements-feedback-sKrDyQxcfA2">feedback</a> will help us to improve it.</span>
+    <span>This is a new service – your <a href="#">feedback</a> will help us to improve it.</span>
   </p>
 </div>
--- a/app/views/guide_alpha_beta.html
+++ b/app/views/guide_alpha_beta.html
@@ -94,8 +94,8 @@
   </p>
 
   <p>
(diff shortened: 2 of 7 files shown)


igloosi:
cc @timpaul

app/views/includes/phase_banner.html
```diff
@@ -1,6 +1,6 @@
 <div class="phase-banner">
   <p>
     <strong class="phase-tag">ALPHA</strong>
-    <span>This is a new service – your <a href="https://designpatterns.hackpad.com/GOV.UK-elements-feedback-sKrDyQxcfA2">feedback</a> will help us to improve it.</span>
```
selfthinker: I don't think this is correct. The original link went to something about feeding back in general, not specifically for the phase banners. As this include isn't used anywhere right now, I would probably just exchange the link with a `#` so that a new link can be added whenever more feedback is required.

selfthinker:
Minor: Most of the links (which Tim added in the referenced issue) are sort of redirects, but not all of them. (Dropbox Paper adds the page title into the URL after a couple of seconds.) Should we use only the redirects or only the links to which they redirect?

igloosi:
update to links no non-redirect versions

app/views/includes/phase_banner.html
igloosi: replaced with #

Issue #536, opened 2017-07-21
Change to type=tel for phone numbers


edwardhorsford:
We should change to `type=tel` for phone numbers in Elements examples - this should bring up the correct keypad on mobile devices and potentially help with autofill.

NB - will obviously want to test this a bit, but it feels like a sensible change.

36degrees:
Fixed in #607.

PR #607, opened 2018-05-31
Use `<input type=“tel”>` for mobile keyboard

flexbox:
For phone number the keyboard on mobile is the text one
05_31_18 at 09 50 46am

with xCode ios simulator

05_31_18 at 10 33 02am


robinwhittleton:
I think @edwardhorsford has some opinions on this.

edwardhorsford:
My only opinion is that we should do this ;)

I raised this as an [issue](https://github.com/alphagov/govuk_elements/issues/536) last year, so glad to see it addressed.

PR #611, opened 2018-07-04
Fix missing arrow on `<details>` element In Firefox

igloosi:
Removes the Firefox hack and replace it with the custom css triangle solution developed in GOV.UK Frontend.
Also deletes the _shame.scss file as it's no longer needed.

Fixes a missing open/close indicator on `details` element in Firefox.

Firefox:
screen shot 2018-07-04 at 15 37 21
screen shot 2018-07-04 at 15 37 40

Other browsers:
Preview url: https://govuk-elements-review-pr-611.herokuapp.com/typography/#typography-hidden-text


36degrees:
Seeing duplicated arrows in IE8 through IE11. I believe this is because these browsers do not natively support the `<details>` element, and we add the marker using JavaScript. We don't have this problem in GOV.UK Frontend because we've removed that responsibility from the JS.

36degrees:
Annoyingly, that happens over in https://github.com/alphagov/govuk_frontend_toolkit/blob/master/javascripts/govuk/details.polyfill.js#L174-L186

igloosi:
yeah, just fixed that

PR #621, opened 2019-06-27
Update deprecation_message.html

amyhupe:
Updating the deprecation notice to include information about accessibility updates.

### DIFF
--- a/app/views/includes/deprecation_message.html
+++ b/app/views/includes/deprecation_message.html
@@ -1,5 +1,5 @@
 <div class="notification-summary">
-  <h2 class="heading-medium no-top-margin">The GOV.UK Design System launched on 22 June 2018</h2>
-  <p>The Government Digital Service (GDS) is retiring GOV.UK Elements following the launch of the GOV.UK Design System. GOV.UK Elements will remain available to you if you are currently using it but will no longer be maintained. GDS will only roll out major bug fixes and security patches.</p>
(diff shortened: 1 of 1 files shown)


kr8n3r:
can we also update the message in the README file?

app/views/includes/deprecation_message.html
```diff
@@ -1,5 +1,5 @@
 <div class="notification-summary">
-  <h2 class="heading-medium no-top-margin">The GOV.UK Design System launched on 22 June 2018</h2>
-  <p>The Government Digital Service (GDS) is retiring GOV.UK Elements following the launch of the GOV.UK Design System. GOV.UK Elements will remain available to you if you are currently using it but will no longer be maintained. GDS will only roll out major bug fixes and security patches.</p>
-  <p>For new projects, you can get all of the information about GOV.UK styles, components and patterns in the <a href="https://www.gov.uk/design-system">GOV.UK Design System</a>.</p>
+<h2 class="heading-medium no-top-margin">GOV.UK Elements has now been replaced by the GOV.UK Design System</h2>
+<p>Elements will remain available in case you are currently using it, but will no longer be maintained. The Government Digital Service will only carry major bug fixes and security patches.</p>
```
nickcolley: Given that we're now saying 'it has now been replaced'

Could we change 'but will no longer be maintained' to 'is no longer maintained'.

PR #628, opened 2019-12-20
Remove cookie banner

nickcolley:
When we merge https://github.com/alphagov/govuk_elements/pull/627 we'll not set any cookies so we can remove the cookie banner.

I've had to do some undesirable hot patching to do this due to the limitations of govuk_template.

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -2,9 +2,4 @@
 
 // Note: this config can be overridden using environment variables (eg on heroku)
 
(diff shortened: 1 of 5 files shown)


nickcolley:
Another approach that others have tried is to immediately clear the 'seen' cookie, alongside hiding it with CSS: https://github.com/alphagov/datagovuk_find/pull/714

PR #629, opened 2020-04-21
Update GOV.UK Elements deprecation notices

m-green:
We’re updating the deprecation notice in the Elements, Template and Frontend Toolkit repos, to reflect the fact that GOV.UK Design System and Frontend have now been updated to better meet WCAG 2.1.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,13 @@
 The GOV.UK Design System launched on 22 June 2018
 ===============
 
-GOV.UK Elements has now been replaced by the GOV.UK Design System. Elements will remain available in case you are currently using it, but is no longer maintained. The Government Digital Service will only carry out major bug fixes and security patches.
+GOV.UK Elements is:
 
-The GOV.UK Design System will be updated to ensure the things it contains meet level AA of WCAG 2.1, but Elements will not. [Read more about accessibility of the GOV.UK Design System](https://design-system.service.gov.uk/accessibility/).
+- no longer maintained
+- will only be updated for major bug fixes and security patches
+- does not meet the [Web Content Accessibility Guidelines (WCAG 2.1 level AA) accessibility standard](https://www.gov.uk/guidance/accessibility-requirements-for-public-sector-websites-and-apps)
 
+This framework will remain available in case you’re currently using it. To help your service meet accessibility requirements, you should use the [GOV.UK Design System](https://design-system.service.gov.uk/). You can [migrate to the Design System from GOV.UK Elements](https://frontend.design-system.service.gov.uk/migrating-from-legacy-products/).
 
 GOV.UK elements ·
 [![JavaScript Style Guide](https://img.shields.io/badge/code_style-standard-brightgreen.svg)](https://standardjs.com)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,10 +1,13 @@
 The GOV.UK Design System launched on 22 June 2018
 ===============
 
-GOV.UK Elements has now been replaced by the GOV.UK Design System. Elements will remain available in case you are currently using it, but is no longer maintained. The Government Digital Service will only carry out major bug fixes and security patches.
+GOV.UK Elements is:
 
-The GOV.UK Design System will be updated to ensure the things it contains meet level AA of WCAG 2.1, but Elements will not. [Read more about accessibility of the GOV.UK Design System](https://design-system.service.gov.uk/accessibility/).
+- no longer maintained
+- will only be updated for major bug fixes and security patches
+- does not meet the [Web Content Accessibility Guidelines (WCAG 2.1 level AA) accessibility standard](https://www.gov.uk/guidance/accessibility-requirements-for-public-sector-websites-and-apps)
```
hannalaakso, line 8: Sorry I didn't spot this earlier but I'm wondering whether the link text here could be more in line with the page title of the page it links to, or could link to [Meeting accessibility requirements](https://www.gov.uk/guidance/accessibility-requirements-for-public-sector-websites-and-apps#Meeting%20accessibility%20requirements).

app/views/includes/deprecation_message.html
```diff
@@ -1,5 +1,14 @@
 <div class="notification-summary">
 <h2 class="heading-medium no-top-margin">GOV.UK Elements has now been replaced by the GOV.UK Design System</h2>
-<p>Elements will remain available in case you are currently using it, but is no longer maintained. The Government Digital Service will only carry out major bug fixes and security patches.</p>
-<p>The <a href="https://www.gov.uk/design-system">GOV.UK Design System</a> will be updated to ensure the things it contains meet level AA of WCAG 2.1, but Elements will not. <a href="https://design-system.service.gov.uk/accessibility/">Read more about accessibility of the GOV.UK Design System</a>.</p>
+
+<p>GOV.UK Elements is:<p>
+
+<ul>
```
hannalaakso, line 6: This will need the [bullet list style ](<ul class="list list-bullet">
 ) applied. Once that's added, the list will have 45px bottom margin (from `.list-bullet`) which makes it look a bit unbalanced.

Screenshot 2020-04-22 at 15 39 33

If we want to fix that, we'll need to add a new class [Elements documentation Sass](https://github.com/alphagov/govuk_elements/blob/master/assets/sass/elements-documentation.scss) as Elements doesn't have spacing override classes. I've just gone through the Elements sass files but I can't see another class we could use here to remove the bottom margin.

README.md
m-green: No worries - I've updated the link in this and the other PRs.

app/views/includes/deprecation_message.html
m-green: Thanks for spotting - I've added the CSS classes - let me know if I've done it correctly. Are you happy to leave the margin as is? Feels like it'd probably be going too far to update Elements just for this margin.

36degrees: I think I'd agree – or, if we think it's important, can we just add an inline style to the `ul`?

hannalaakso: @m-green Happy for you to leave it as is, or like @36degrees suggested you can add `style="margin-bottom: 20px;"` to the `ul` element.

PR #635, opened 2020-04-30
Fix security vulnerabilities (Gulp v4, cssnano, gulp-mocha v7 upgrades)

hannalaakso:
There are currently multiple security vulnerabilities reported for npm packages:

```
found 12 vulnerabilities (3 low, 1 moderate, 7 high, 1 critical) in 10514 scanned packages
9 vulnerabilities require semver-major dependency updates.
3 vulnerabilities require manual review. See the full report for details.
```

To resolve the alerts, this PR:
- Updates from Gulp 3.91 to Gulp 4.0.2 (latest) 
- Updates project Gulp tasks to use Gulp 4 (with some minor tidy up)
- Replaces deprecated `gulp-cssnano` with `cssnano` via `gulp-postcss`
- Updates `gulp-mocha` three major versions up (4.3.1 -> 7.0.2)

Recommend reviewing the commits individually.

- `npm test` passes and fails automated tests appropriately
- `npm start` removes the `/public` folder, outputs both a minified and non-minified version of CSS into `/public/stylesheets`, copies JavaScript to `/public/javascripts`, copies images to `/public/images` and starts the nodemon server
- 'cssnano' with 'gulp-postcsss' (which replace deprecated 'gulp-cssnano') minify `/public/stylesheets/govuk-elements-styles.min.css` file correctly
- `gulp package` generates `/package/` with changes for publishing
- `gulp default` outputs the available main tasks correctly

- Fixing security vulnerabilities

- My change requires no change to the documentation.

Fixes https://github.com/alphagov/design-system-team-internal/issues/285


gulpfile.js
```diff
@@ -58,9 +57,10 @@ gulp.task('scripts', () => {
 // Runs tasks that copy assets to the public directory.
 // ---------------------------------------
 
-gulp.task('build', cb => {
-  runsequence('clean', ['styles', 'images', 'scripts'], cb)
-})
+gulp.task('build', gulp.series(
+  'clean',
+  ['styles', 'images', 'scripts']
```
36degrees, line 63: I can't find anything in gulp's documentation about passing arrays to `gulp.series` – what happens with these tasks? Are they run in series, or parallel?

If they run in series, do they need to be in an array or should this just be `gulp.series('clean', 'styles', 'images', 'scripts')`?

If they run in parallel, can we make that explicit by doing `gulp.series('clean', gulp.parallel('styles', 'images', 'scripts'))`?

(I may just be missing something here, `gulp` is not one of my strong points!)

gulpfile.js
```diff
@@ -84,46 +84,35 @@ gulp.task('server', () => {
 
 gulp.task('test:app', () =>
   gulp.src(paths.testSpecs + 'app_spec.js', {read: false})
-  .pipe(mocha({reporter: 'spec'}))
+  .pipe(mocha({reporter: 'spec', exit: true}))
   // https://github.com/sindresorhus/gulp-mocha#test-suite-not-exiting
```
36degrees, line 89: Is this comment still relevant, or should it be removed?

gulpfile.js
```diff
@@ -84,46 +84,35 @@ gulp.task('server', () => {
...
 // Develop task --------------------------
 // Runs copy-assets task and sets up watches.
 // ---------------------------------------
-gulp.task('develop', cb => {
-  runsequence('build',
-              'watch',
-              'server', cb)
-})
+
+gulp.task('develop', gulp.series(
```
36degrees, line 112: Minor, but if I `Ctrl-C` this task I get:

```
[09:30:33] The following tasks did not complete: develop, server
[09:30:33] Did you forget to signal async completion?
```

This doesn't happen on `master`.

gulpfile.js
```diff
@@ -141,9 +130,9 @@ gulp.task('package:prepare', () => {
     .pipe(gulp.dest(paths.package + 'public/sass/'))
 })
 
-gulp.task('package', cb => {
-  runsequence('package:prepare', cb)
-})
+gulp.task('package', gulp.series(
```
36degrees: I don't think this is something that needs fixing in this PR, but because this task doesn't delete the contents of the `package/public/sass` directory, files don't get deleted from the package when they're deleted from `/assets/sass` – so  [`_shame.scss`](https://github.com/alphagov/govuk_elements/blob/77783b1dfecb74048f8824a720aa8da9d9cdbba5/packages/govuk-elements-sass/public/sass/elements/_shame.scss) is still in the package folder, despite having been deleted in #611.

gulpfile.js
hannalaakso: That's a good point, I've made it more explicit with `gulp.parallel`.

gulpfile.js
hannalaakso: It was still relevant as were using something from that part of the docs. However I've taken it out as we were essentially just linking to the docs for the library itself which seems redundant.

gulpfile.js
hannalaakso: Thanks, fixed now.

gulpfile.js
hannalaakso: I've had a go at resolving this in https://github.com/alphagov/govuk_elements/pull/635/commits/93ffb917ee66f52c4e0f516c48a18cb5806119dd. But I could split it into a separate PR? (If we keep it here I'll remove https://github.com/alphagov/govuk_elements/pull/635/commits/9993eee8af6eb9d60228ac828e63ec4aea016620)

36degrees: Yep, let's split this out and worry about it later – I don't think that's quite right as a lot of `/package` is not generated from anywhere (e.g. `/package/package.json`)

hannalaakso: Yes that's the reason only `packages/govuk-elements-sass/public` is removed and the files at `/packages/` level aren't touched. I'll raise the PR once this is merged.

36degrees: Ah, sorry, I think I was thrown by the commit message. Let's review it separately 👍

hannalaakso: Yeah the commit message could have been clearer. Cool I'll raise that PR separately.

PR #637, opened 2021-09-14
Remove Travis configuration

lfdebrux:
We are unlikely to do another release from this repo, and we don't want to have to keep a load of secrets in Travis just in case we did want to.

This PR removes all the Travis configuration, and updates the CONTRIBUTING file to outline how to do a release without Travis.

It also adds a simple PR check using GitHub Actions, just in case.


36degrees:
I have disabled GitHub Actions in this repo until we work out what to do with this – if we pick this work back up we should re-enable Actions whilst ensuring that we're following best practises.

